PR #19, opened 2016-07-14
Develop

dithillobothrium:


dithillobothrium:
Maybe one need to check parallelization of Band::solve_for_kset after I moved here this code from K_set::find_eigen_states

dithillobothrium:
Maybe one need to check parallelization of Band::solve_for_kset after I moved here this code from K_set::find_eigen_states. But it seems to be working well as before

PR #144, opened 2018-01-30
get rid of pp_desc structure in atom type

toxa81:
@mtaillefumier : I commented angular_momentum and total_angular_momentum in radial integrals of pseudo wave-functions. Please, confirm, that it doesn't break any of your code.


toxa81:
@mtaillefumier : the pull request is large but it should not, in principle, touch your code. But please, check it.

PR #188, opened 2018-03-28
Develop

ckae95:


apps/dft_loop/sirius.scf.cpp
```diff
@@ -254,6 +254,7 @@ void run_tasks(cmd_args const& args)
         ks.add_kpoint(&vertex[0].second[0], 1.0);
 
         double t{0};
+        std::cout<<"vertex.size = " << vertex.size();
```
toxa81: this has to be removed

toxa81:
* apps/dft_loop/timers.json has to be removed
* all libraries, tar.gz archives and binary files have to be removed (and not even tracked)

PR #223, opened 2018-06-14
Develop

mtaillefumier:
Enclosed is the latest modifications I have done on SIRIUS. they include
- multigpu support
- first version of the hubbard forces and stresses (but it is incomplete)

There are a couple of things that might bite us back on the multigpu support. First we are limited to a small number of cuda stream (32 per card) which means that we have to be extremely careful with the openmp num threads and the number of mpi ranks. Second if we run for instance 4 ranks with two cards, we need to set up the OMP_NUM_THREADS accordingly so that no more than 32 streams are created on each card. OMP_NUM_THREADS=8 and mpi_ranks=4 might work or not. Above that we will most certainly have deadlock in the cuda libraries.

I have tested the multi gpu support and it works fine provided we stay within boundaries. However QE does not like the power boxes and can not process its own input file on them. I will recompile the all thing on greina to check that QE does not get crazy when multigpu is used.

We will need to do a proper reformating of the code to remove all trailing whitespaces at least in SIRIUS. I have the necessary configuration for vim as well so you do not need to think about it. And I have git hook to fix that definitely. Unfortunately we can not impose it for QE but we can always create a patch excluding whitespace modification when necessary
M.


src/Geometry/force.hpp
```diff
@@ -799,6 +792,11 @@ class Force
                                forces(0, ia), forces(1, ia), forces(2, ia));
                     }
                 };
+            if (ctx_.gamma_point() && (ctx_.so_correction() == false)) {
```
electronic-structure: braces {} of the previous lambda are at the wrong identation

src/Geometry/stress.hpp
```diff
@@ -931,30 +932,31 @@ class Stress {
...
                 }
             }
-            kp__->beta_projectors().dismiss();
-            bp_strain_deriv.dismiss();
-            hamiltonian_.dismiss();
         }
 
+        hamiltonian_.dismiss();
+
         // global reduction
```
electronic-structure: /* comments in code */

src/Geometry/stress.hpp
```diff
@@ -931,30 +932,31 @@ class Stress {
...
         ctx_.comm().allreduce<double, mpi_op_t::sum>(&stress_hubbard_(0, 0), 9);
```
electronic-structure: ctx_.comm().allreduce(&stress_hubbard_(0, 0), 9);
also works

src/Geometry/force.hpp
```diff
@@ -809,6 +807,7 @@ class Force
 
                 hubbard_force_add_k_contribution_colinear(*kp, forces_hubbard_);
             }
+            hamiltonian_.dismiss();
 
             print_forces(forces_hubbard());
             // global reduction
```
electronic-structure: /* comments in code */

src/Hamiltonian/non_local_operator.hpp
```diff
@@ -291,6 +291,11 @@ inline void Non_local_operator<double>::apply(int chunk__,
         }
     }
 
+#ifdef __GPU
+    if (pu_ == GPU)
```
electronic-structure: if (pu_ == GPU) {
    assert(...);
}

mtaillefumier:
fixed good to go

mtaillefumier:
It should be fine.

mtaillefumier:
by the way, you have the possibility to hide whitespace changes when you open a commit. Just found this option.

PR #285, opened 2018-10-09
Feature/robust optimization

simonpintarelli:
implementation of the orbital transformation method (10.1063/1.1543154) in Python.

small changes / bug fixes in cpp code:
- Local_operator::apply, avoid overwriting other ispn with zeros
- avoid illegal memory access in pp/exact diagonalization (results are still wrong)
- omit Field4D::allocate in pp case
- make UPF converter scripts more robust (omit size checks if length field in UPF file is missing)
- warn about negative densities only if above threshold of 1e-9
- fix test_eigen_v2.cpp


toxa81:
I can't merge it, github tells 
"Couldn’t update branch
Oops, something went wrong"

electronic-structure:
can you close this one and reopen it again?

PR #311, opened 2018-12-07
Memory and linalg refactoring

toxa81:
Refactoring of memory management and linear algebra

The primary goal of this modifications is to introduce a new interface
to the linear algebra routines, i.e. be able to call BLAS routines
like this:
```cpp
  auto la = linalg_t::blas;
  linalg2(la).zgemm('N', 'N', m, n, k, a.at(memory_t::host), a.ld()...)
```
The following changes were implemented

* `memory_pool` class was finalized and benchmarked; the last thing to implement in memory_pool is a proper memory alignment, i.e. return pointers that are compliant with alignof(T) length
* `mdarray` class is in the transition state:
  * new method `.at(memory_t, i, j..)` is introduced; templated version `.at<CPU>(i, j...)` is going to be removed in the future
  * new method `.copy_to(memory_t..)` is introduced; old templated version `.copy<memory_t from, memory_t to>()` will be removed in the future
  * in general, the usage of templates will be removed where it is not justified
  * `memory_pool` and `mdarray` classes are now located in the same file memory.hpp as they both are used to manage the memory
  * `mdarray` can now allocate from the memory pool
* new linalg2 interface: now cublasxt with CPU pointers can be called
* reduction of GPU memory usage in `Local_operator::apply_h()`
* minor text formatting in Hubbard-U descriptor class
* remove template parameter N in beta_projector_base class; same reasoning: template parameter is not justified
* new inner(), transform() and ortho() version that accept memory type and linear algebra type instead of processing device; old version will be removed in the future


mtaillefumier:
Seems to be fine after going through the changes. We will need to restore the set_device thingy when this is merged. There is nothing wrong with the commit a64bf3af45092d416ee9239a7a82b8fb9b95f317. I think we can merge it and do more testing.

By the way, I think it would be good to remove commented dead code when relevant afterwards.

simonpintarelli:
@toxa81 The python code does not compile anymore. I'll make the required changes.

PR #326, opened 2019-02-08
Mathieu develop

mtaillefumier:
second version of the vdw stuff. I will be preparing verification tests but the implementation was checked with C60 fullerene (crystal and molecule).


mtaillefumier:
Note : vdwxc is not a hard dependency.

PR #352, opened 2019-04-10
GGA formula, reduce copies in H apply

simonpintarelli:
- improve GGA (magnetic case)
- copy only required components in sphi (S*phi) during `Hamiltonian::apply_h_s`


toxa81:
I think this is due to small change in XC energy for magnetic systems

PR #355, opened 2019-04-18
use div(grad rho vsigma) in the MT 

toxa81:


toxa81:
new test17 doesn't pass; the total energy is slightly different from what I get on my laptop

simonpintarelli:
test17 is passing on my desktop too

PR #357, opened 2019-04-24
Python / Cmake / minor changes

simonpintarelli:
- update python
- ifdef for libvdw depending on MPI support
- remove prev_rms / rms_avg cout from mixer
- call `CudaProfilerStop` in finalize
- assume `paw_core_energy=0` if there is no corresponding entry in the UPF file
- add `apply_s` (overlap only) function to Hamiltonian
- reduce `eps` (1e-9) in `reduce_coordinates`


toxa81:
@simonpintarelli if it's ok to merge it, please, go ahead

PR #358, opened 2019-04-26
Expose pw coeffs in density to python

simonpintarelli:


src/field4d.hpp
```diff
@@ -147,24 +147,37 @@ class Field4D
     Periodic_function<double>& vector(int i)
     {
         assert(i >= 0 && i <= 2);
+        if (components_[i+1] == nullptr) {
```
toxa81: Is this performance optimal? I just don't know. If the main code has something like
for (int ig = 0; ig < ng; ig++) {
  density.component(0).f_pw_loca(ig)
...
}
will the 'if` statement here be checked on each iteration of ig?

toxa81: suggestion to switch to assert()

src/smooth_periodic_function.hpp
```diff
@@ -119,6 +119,11 @@ class Smooth_periodic_function
         return f_pw_local_(ig__);
     }
 
+    inline mdarray<double_complex, 1>& pw_array()
```
toxa81: it's better to rename it to the variable: f_pw_local()

src/field4d.hpp
simonpintarelli: I don't know if the compiler would move component(0) outside the for loop, perhaps if density is const. I did add it because python in this case won't die because of the illegal memory access.
You are right, such for loops appear, I'll remove the if statement.

PR #363, opened 2019-05-03
Misc fixes

dev-zero:
While RedHat seems to install the `spglib` header into its own subdirectory, `spglib` does not seem to do that by itself when simply using `make install`. Distros (including Spack) which follow upstream would currently have to patch SIRIUS to locate the `spglib.h` properly.

The second issue occurs when trying to link dynamically against `libsirius_f.so` (obtain with with CMake's `-DBUILD_SHARED_LIBS=ON`): at this point the linker complains about undefined symbols due to missing linker flags when building the library.

This information will also be needed if you at some point should provide CMake package config/version files for discovering SIRIUS (and the required flags in case of static linking) via CMake (via the CMake functions `configure_package_config_file` and `write_basic_package_version_file`, see for example [DBCSR](https://github.com/cp2k/dbcsr/blob/develop/src/CMakeLists.txt#L284-L300))

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -13,7 +13,7 @@ if(USE_CUDA)
 endif()
 if(CREATE_FORTRAN_BINDINGS)
   add_library(sirius_f "sirius_api.cpp;sirius.f90")
-  add_dependencies(sirius_f generate_version_hpp runtime_options_json_hpp)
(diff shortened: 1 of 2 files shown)


simonpintarelli:
Thank you! We have merged your commits into the develop branch (#364 and #365).

PR #375, opened 2019-06-11
added a check of S-matrix

toxa81:
With --control.verification=1 all the 17 tests pass on my laptop. This verification level enables the check of the S-matrix. For bad systems (where orthogonalziation fails) the check fails as well.


simonpintarelli:
retest this please

simonpintarelli:
merging (the failed test is noise from jenkins).

PR #377, opened 2019-06-13
updates to reframe; redundant code removed

toxa81:
@simonpintarelli : I removed apply_s and used apply_h_s instead. This might impact the python module. Can you check that your code is still working.


simonpintarelli:
@toxa81 Thanks. This looks fine. I don't know why I've added `apply_s`.

PR #382, opened 2019-07-09
CMake: move cuda into `libsirius.a`, fotran mod file

simonpintarelli:
- install Fortran mod file into `include/sirius`
- build only one static library: `libsirius.a`


toxa81:
I will merge it and work on possible issues

PR #395, opened 2019-07-22
Split cpp hpp

AdhocMan:
Declarations and definitions split in folders sddk and utils.


simonpintarelli:
cherry-picked commits already. Closing PR.

PR #407, opened 2019-09-13
SpFFT and not only

toxa81:
@simonpintarelli @mtaillefumier 
Guys!
I must apologise for this gigantic pull request. Things went beyond my control :)
So, what's inside?
1. A full dependency on the SpFFT library.
FFT3D class is gone. This requires the updates of the Jenkins scripts, such that SpFFT is loaded before. To build SIRIUS you need to pass `-DSpFFT_DIR=/path/to/spfft/lib/cmake/`
2. A separation of Hamiltonian class into k-independent and k-dependent parts
Now the hamiltonian works like this:
```
Hamiltonian H0(potential)
auto Hk = H0(kp)
Hk.apply_h_s(....)
```
all the initial preparations of the potential on a coarse mesh and beta-projectors happen inside the constructors. 
3. Some refactoring of the U-part
Generation and orthogonalization of the Hubbard wave functions moved to the K-point class. Application of the S-operator is converted into a pure function. Hubbard object is moved from Hamiltonian to Potential class. 

There are several things that need to be fixed/improved before we merge the request. Jenkins has to be aware of SpFFT (we have EB module but it is not pushed to production repository yet). Items 1,2 broke the Python frontend, @simonpintarelli can you please have a look and fix what is needed. Item 3 broke the allocation of Hubbard wave-functions of GPUS. Is was done in the `generate_atomic_orbitals` and now it is not. @mtaillefumier can you please check where the Hubbard wave-functions are used and allocate memory explicitly (like we do for normal wave-functions).


toxa81:
It actually has to be merged to 'develop'

simonpintarelli:
@toxa81 I'll take care of the Python bindings and jenkins, the EB modules are installed in my home directory on daint.

toxa81:
It looks better now. The GPU tests are failing, but this is the LAPW part, which I'll take care of.

toxa81:
All 18 tests pass on CPU sequential. GPU sequential is broken, looking into it

toxa81:
Looks that all tests pass!

toxa81:
It was a pinned memory problem. Now GPU code with scalapack works fast

toxa81:
All tests pass!

toxa81:
Working on the fix

PR #412, opened 2019-09-19
Disactivated the mpi version of libvdwxc. This needs proper fixing in…

mtaillefumier:
We never compile SIRIUS with libvdwxc present by default but CP2K does. libvdwxc supports fftw-mpi but there is currently no way to know at compilation time (SIRIUS) which means I can not use their constants to know about it.
This problem can be solved on our side by checking the presence of vdwxc_init_mpi in the library or by fixing the library itself. Right now I simply deactivate the mpi support in libvdwxc


simonpintarelli:
Has been fixed in develop already. CMake will now check whether libvdwxc has been built with mpi.

PR #420, opened 2019-10-04
[WIP] LDA+U

toxa81:
@mtaillefumier, I added a new test19 for the NiO with LDA+U, can you please check it. I don't thinks it's correct, because in LDA+U for NiO we should obtain a gap of 4 eV and magnetic moment of ~1.8 mB. Also, the QE+SIRIUS with LDA+U is broken as it gives different results from run to run.


mtaillefumier:
there is a mistake in the input file. you wrote 
hubard instead of hubbard.

mtaillefumier:
With the syntax error corrected I get this
Charges and magnetic moments
--------------------------------------------------------------------------------
atom              moment                |moment|
--------------------------------------------------------------------------------
   0  [  0.0000,   0.0000,   1.8212]    1.821166
   1  [  0.0000,   0.0000,  -1.8212]    1.821166
   2  [  0.0000,   0.0000,  -0.0018]    0.001849
   3  [  0.0000,   0.0000,   0.0018]    0.001849

total charge          :  48.000000
total moment          : [  0.0000,   0.0000,   0.0000], magnitude :   0.000000

Energy
--------------------------------------------------------------------------------
valence_eval_sum          :       -26.05012938
<rho|V^{XC}>              :       -53.19686760
<rho|E^{XC}>              :       -42.24303981
<mag|B^{XC}>              :        -0.18569835
<rho|V^{H}>               :       243.78412462
one-electron contribution :      -216.63738639 (Ha),      -433.27477279 (Ry)
hartree contribution      :       121.89206231
xc contribution           :       -42.24303981
ewald contribution        :      -238.91986830
PAW contribution          :         0.00000000
Hubbard energy            :         0.84715651 (Ha),         1.69431302 (Ry)
Total energy              :      -374.87537733 (Ha),      -749.75075466 (Ry)

band gap (eV) :         2.41936039
Efermi        :         0.53460745

but it is difficult to converge.

toxa81:
I'm checking the code and I don't see the mixing of the LDA+U occupancy matrix. I think that is the root of the slow convergence.

mtaillefumier:
I implemented the mixing but I do not recall why we removed it. 

I have a fundamental problem with the mixing of the hubbard occupancies matrix. It is fine for the density because we recompute everything with the resulting density but we do not know how to compute the occupancy matrix from the density directly. Basically we do not know how to compute U_Hubbard(n) and mixing does not help much at the beginning. This is an issue with the theory itself since hubbard can not be derived from the Kohn-Sham equations.

PR #431, opened 2019-10-27
cuda 10: get rid of deprecated warning (devicePtr check)

simonpintarelli:
memoryType attribute is deprecated, check if devicePtr is null instead.

### DIFF
--- a/src/SDDK/GPU/acc.hpp
+++ b/src/SDDK/GPU/acc.hpp
@@ -487,7 +487,7 @@ inline bool check_device_ptr(void const* ptr__)
     if (error != cudaSuccess) {
         return false;
     }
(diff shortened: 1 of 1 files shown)


toxa81:
is it compatible with CUDA9?

simonpintarelli:
I think so. According to the v9.1 documentation the devicePtr is null if it cannot be accessed by the current device: [docs.nvidia.com: cudaPointerGetAttributes](https://docs.nvidia.com/cuda/archive/9.1/cuda-runtime-api/group__CUDART__UNIFIED.html#group__CUDART__UNIFIED_1gd89830e17d399c064a2f3c3fa8bb4390). Same for Cuda 10.

PR #446, opened 2019-11-30
printf

simonpintarelli:
- `printf` -> `std::printf` (use C++ printf, touches many files, not necessary)
- respect `-Werror=format-security`


toxa81:
@simonpintarelli  was this PR merged?

simonpintarelli:
@toxa81 Not yet. We first have to decide about the first commit changing `printf` -> `std::printf`. It changes many files  and is not necessary to fix the compilation error.

toxa81:
Looks good to me. We should merge it.

PR #451, opened 2019-12-19
Add extra functions in the api for kpoints information

mtaillefumier:
this pull request includes some requests from cp2k people. Basically, it adds few functions to the api that return properties about the k-point, their number, weight and coordinates.


mtaillefumier:
We will need to create a new minor version

PR #467, opened 2020-01-24
Build SIRIUS as a shared library and refactor CMake

teonnik:
- Build SIRIUS as a shared library by default
- Include RPATHs when installing
- Clean-up CMake scripts


simonpintarelli:
Thank you very much @teonnik! Looks good to me. 

`timer.hpp` had static variables in inlined member functions which caused a segfault when calling the finalize routine. This seems to have surfaced only now after linking against shared libsirius. It is only the python module which seems to be affected though. Fixes are in branch [cmake-rewrite](/electronic-structure/SIRIUS/tree/cmake-rewrite)

At least Debian packagers do not like rpath: 
https://wiki.debian.org/RpathIssue
This seems not to be an issue since all rpath settings can be disabled by setting `CMAKE_SKIP_RPATH` if I understand correctly:
https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling#no-rpath-at-all

teonnik:
Regarding Debian, the current implementation does not embed RPATHs when SIRIUS is installed to system directories. I think that mostly addresses their concerns. 

To the first point made in the Debian link you provided:

```
A problem arises when binary A defines a NEEDED dependency on libraries B.so.1 and C.so.2, while library B.so.1 depends on library C.so.1. This means parts of the code will use one version of a library and other parts another. The many versions of a library scenario is needed to deal with gradual migrations, but maintainers shuffle libraries around when dealing with such a situation and packages with RPATH could end up finding the wrong version of a dependent library, one with incompatible dependencies. 
```

This shouldn't be an issue with a package manager as only one version of a package is allowed in the dependency tree. That's one of it's core responsibilities.

Some more thoughts about RUNPATH and RPATH here: https://github.com/spack/spack/issues/9160#issuecomment-420192189 .

It seems to come down to this: `We like RPATH because we trust the builder of the application, much more than the user, to know the dependencies.`

simonpintarelli:
The binaries built on archlinux have RUNPATH set, not RPATH. On ault RPATH is set as expected.

teonnik:
I made final improvements to the RPATH support. The install folder is now relocatable as binaries in bin use relative paths to libsirius.so. That means that the folder can be moved / renamed without breaking the binaries. 

The same can be done with binaries in the build directory. By default, CMake uses absolute paths, however, since CMake 3.14 the variable `CMAKE_BUILD_RPATH_USE_ORIGIN` makes them relative and the thus the build directory can be moved around / renamed (i.e. relocatable). I left the variable commented out as bumping the version to 3.14 is a point of discussion.

Absolute paths are embedded for dependencies outside of the build/install tree. These are controlled by `CMAKE_INSTALL_RPATH_USE_LINK_PATH`. An option can be introduced for setting this OFF in case only the relative paths (for in-tree binaries) are wanted. I am not sure if there is a scenario where that will be desirable (perhaps shipping pre-built binaries to a remote machine where paths of external dependencies are different?).

teonnik:
There is `CMAKE_SKIP_INSTALL_RPATH` which removes all installation RPATHs/RUNPATHs.

PR #470, opened 2020-01-27
Generate runtime_options_json.hpp with CMake

teonnik:
- There is no need for an intermediary, the json string can be defined
  as a raw string directly fom the hpp file.
- Remove corresponding bash script and CMake targets


AdhocMan:
The generated header probably has to be added to the install command as well

teonnik:
This is ready for review.

PR #473, opened 2020-01-27
save the branch name in SIRIUS_GIT_BRANCH

toxa81:
No hash in case of release archive.


toxa81:
@teonnik One last issue that we need to pin is the absence of 'git' command. Yes, this might be the case when user checkin code on one node and then copy the whole tree to another place where 'git' is not installed.

teonnik:
To check if the `git` commands is found : `find_program(GIT_EXECUTABLE NAMES git)` and
then `if(DEFINED GIT_EXECUTABLE AND EXISTS .git)`.

PR #474, opened 2020-01-27
[cmake] fix relative rpath for python module

simonpintarelli:
- relative RPATH was wrong for the Python module

### DIFF
--- a/python_module/CMakeLists.txt
+++ b/python_module/CMakeLists.txt
@@ -29,20 +29,28 @@ if(CREATE_PYTHON_MODULE)
     target_link_libraries(${libname} PRIVATE ${MKL_DEF_LIBRARY})
   endif()
 
+  set(PYTHON_INSTALL_SITE_DIR ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages)
+
+  # set relative RPATH
+  if(isSystemDir STREQUAL "-1")
+    file(RELATIVE_PATH relDir ${PYTHON_INSTALL_SITE_DIR}/sirius
+      ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR})
+    set_target_properties(${libname} PROPERTIES INSTALL_RPATH "${basePoint};${basePoint}/${relDir}")
+  endif()
+
   target_link_libraries(${libname} PRIVATE sirius)
   # collect python files in module dir
   # install to cmake prefix
-
   if(NOT PYTHON2)
     install(DIRECTORY sirius
       DESTINATION
-      ${CMAKE_INSTALL_PREFIX}/lib/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages/
+      ${PYTHON_INSTALL_SITE_DIR}
       FILES_MATCHING REGEX
       ".*py"
       )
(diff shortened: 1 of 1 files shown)


python_module/CMakeLists.txt
```diff
@@ -29,20 +29,37 @@ if(CREATE_PYTHON_MODULE)
...
+  set(PYTHON_INSTALL_SITE_DIR ${CMAKE_INSTALL_PREFIX}/lib/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages)
+
+  # set relative RPATH
+  if(isSystemDir STREQUAL "-1")
+    set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+    if(APPLE)
+      set(basePoint @loader_path)
+    else()
+      set(basePoint $ORIGIN)
+    endif()
```
teonnik: Perhaps the basePoint variable can be reused from the top CMake file?

simonpintarelli: makes sense

python_module/CMakeLists.txt
```diff
@@ -29,20 +29,32 @@ if(CREATE_PYTHON_MODULE)
     target_link_libraries(${libname} PRIVATE ${MKL_DEF_LIBRARY})
   endif()
 
+  set(PYTHON_INSTALL_SITE_DIR ${CMAKE_INSTALL_PREFIX}/lib/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages)
+
+  # set relative RPATH
+  if(isSystemDir STREQUAL "-1")
+    set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
```
teonnik: This is already set in the top CMake file.

python_module/CMakeLists.txt
```diff
@@ -29,20 +29,32 @@ if(CREATE_PYTHON_MODULE)
...
+    file(RELATIVE_PATH relDir ${PYTHON_INSTALL_SITE_DIR}/sirius
+      ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR})
+    message("reldir in python: ${relDir}")
```
teonnik: Perhaps this was only needed for debugging?

python_module/CMakeLists.txt
```diff
@@ -29,20 +29,32 @@ if(CREATE_PYTHON_MODULE)
     target_link_libraries(${libname} PRIVATE ${MKL_DEF_LIBRARY})
   endif()
 
+  set(PYTHON_INSTALL_SITE_DIR ${CMAKE_INSTALL_PREFIX}/lib/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages)
```
teonnik: I suggest you use `${CMAKE_INSTALL_LIBDIR}` instead of `lib`, sometimes it's lib64. This can simplify `relDir` to `python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages` and you can just set it directly.

python_module/CMakeLists.txt
simonpintarelli: yes :)

python_module/CMakeLists.txt
simonpintarelli: relDir should be `$ORIGIN/../../../`. I let cmake compute it instead of hardcoding it.

teonnik:
Looks good to me!

PR #484, opened 2020-03-09
Configure Jenkins to push build logs to another repo

haampie:
The idea is to push the build artifacts / logs to another GitHub repo.

TODO:

- [x] configure the bot / user that pushes from Jenkins
- [x] move the repo under this org.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,8 +8,8 @@ pipeline {
     }
     environment {
         EB_CUSTOM_REPOSITORY = '/users/simonpi/jenkins/production/easybuild'
-        LOGS_REPO = 'git@github.com:haampie/SIRIUS_logs.git'
-        LOGS_TREE_URL = 'https://github.com/haampie/SIRIUS_logs/tree/master/'
+        LOGS_REPO = 'git@github.com:electronic-structure/SIRIUS-ci-logs.git'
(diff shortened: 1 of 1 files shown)


haampie:
@simonpintarelli I think I might need push access to this repo in order to be able to test the Jenkins config from the PR itself, otherwise Jenkins discards it as untrusted and uses the old Jenkinsfile.

haampie:
Ok, so it seems there is a security problem when interpolating `SSH_KEY_PATH` as a Groovy variable instead of the shell env variable (ref https://jenkins.io/doc/pipeline/steps/credentials-binding/), should fix that.

A nice feature to use instead of the approach above is `unarchive` s.t. artifacts are specified only once. Unfortunately Jenkins is lacking conditional `post` steps, so for now I would have to wrap the upload & comment feature in a script block.

simonpintarelli:
> @simonpintarelli I think I might need push access to this repo in order to be able to test the Jenkins config from the PR itself, otherwise Jenkins discards it as untrusted and uses the old Jenkinsfile.

@haampie You need write access to the electronic-structure/SIRIUS, otherwise you PR won't be built by jenkins. Seems Anton gave you the permissions already.

haampie:
Ok, thanks, running it again now with the current Jenkinsfile.

haampie:
Seems like the vm runs an ancient version of git :/

PR #485, opened 2020-03-11
update travis ci and front page documentation

toxa81:


toxa81:
Update develop to make a new tag and work on EB scripts.

PR #495, opened 2020-05-07
add robust wave-function optimization interface and other stuff

simonpintarelli:
Sorry, many things in one PR:

- enable for building with clang 
- update python api
- fix issues with static variables in header files (cublas/cusolver handles)
- vdwxc depends on lattice vectors, recreate vdw xc functionals during `Potential::update`
- changes required for robust wave-function optimization, store wave-functions as `std::shared_ptr`
~~- Force OpenBLAS if MKL is not specified, because FindLAPACK will find incomplete MKL if MKL is present on the system but not enabled~~
- gvec_cart replace runtime if-else by std::enable_if
- print some more cublas errors


.github/workflows/build.yml
```diff
@@ -13,7 +13,7 @@ jobs:
           submodules: recursive
 
       - name: Install tools
-        run: sudo apt-get update && sudo apt-get install -y gfortran mpich libmpich-dev liblapack-dev libhdf5-dev libgsl-dev libfftw3-dev libxc-dev
+        run: sudo apt-get update && sudo apt-get install -y gfortran mpich libmpich-dev libpenblas-dev libhdf5-dev libgsl-dev libfftw3-dev libxc-dev
```
toxa81, line 16: libopenblas-dev

simonpintarelli: Thanks for spotting the typo, unfortunately Ubuntu does not package the cmake files or the openblas version is too old: https://packages.ubuntu.com/xenial/amd64/libopenblas-dev/filelist

toxa81:
Now reframe complains. I'll check it today.

mtaillefumier:
In prevision to cp2k integration, would it be possible to add the extra options you need to control the algorithm in the options.json file. A separate commit might be a idea so we can check if there are mistakes.

PR #502, opened 2020-06-08
Multiple fixes

AdhocMan:
Multiple small fixes:

- Removed static qualifier in residuals.cpp for functions with external linkage
- Fixed compilation with CUDA 9 by removing reference to unused fill type
- Fixed compilation with ROCm by disabling reference to CUBLASXT
- Removed redundant use of std::move which prevented copy elision


toxa81:
LGTM

PR #509, opened 2020-06-25
Locking

haampie:
This implements locking of converged vectors at restart. It should reduce the time spend on the dense eigensolver, and should work particularly well when the search subspace size is kept rather small (= more restarts), that is `subspace_size = 2` or so.

It introduces at least one new parameter `block_size`, which is currently hard-coded to be `num_bands / 2`. The idea being that you can do twice as many iterations per restart, s.t. it converges quicker to a subset of the eigenvectors, which can then be locked.

Currently only the smallest eigenpairs are locked. They should be theoretically first to converge anyways. Sometimes it happens though that interior eigenpairs converge first; they are just kept in the search subspace. We could do locking for them, but then potentially we have to permute eigenvectors later again, and potentially we might even need to remove an eigenpair later, when it turns out to have an eigenvalue larger than the `num_bands` smallest ones.

- [x] cuda / cpu version
- [x] flag to enable locking
- [x] implement some missing methods for magma / elpa (probably) for getting eigenvalues from a submatrix
- [x] make it a bit less memory hungry; currently  Hpsi, Spsi and res also allocate num_phi vectors, which is more than necessary, in my julia code only one of these is of size num_phi and the others are num_bands.
- [ ] this one is very bug-prone to implement :sweat_smile:, so I have reverted it at some point: when almost all vectors have converged except for a handful, convergence is very slow. It seems this is because the residuals are near 0 anyways, and the preconditioned residuals are poor quality because of large numerical errors (for instance, the diagonal of the cholesky decomp for interior orthogonalization contains many near 0 values). In my experience it's better to compute a few more residuals of eigenpairs with slightly larger eigenvalues, and converging towards them also improves convergence to the last targeted vecs -- this is because those residuals are not close to 0, so less numerical noise, and they contain components in the direction of smaller eigenvalues as well.
- [x] test it more with `converge_by_energy` enabled
- [ ] implement the same for the other davidson solvers in the SIRIUS code base

### DIFF
--- a/src/band/band.cpp
+++ b/src/band/band.cpp
@@ -44,7 +44,7 @@ Band::Band(Simulation_context& ctx__)
 
 template <typename T>
 void
-Band::set_subspace_mtrx(int N__, int n__, Wave_functions& phi__, Wave_functions& op_phi__, dmatrix<T>& mtrx__,
+Band::set_subspace_mtrx(int N__, int n__, int num_locked, Wave_functions& phi__, Wave_functions& op_phi__, dmatrix<T>& mtrx__,
                         dmatrix<T>* mtrx_old__) const
 {
     PROFILE("sirius::Band::set_subspace_mtrx");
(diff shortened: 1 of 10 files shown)


toxa81:
LGTM

PR #515, opened 2020-08-03
Add spack-based CI

haampie:
The idea is to:
- add a spack repository to this repo
- define some specs / variants of sirius to build as part of CI using `spack dev-build`
- build the spec dependencies upfront inside of a docker image, and store that on dockerhub, using `spack install --only=dependencies` *
- always use the latest repository from spack itself, s.t. we continuously test against new packages down the line

This should make it rather easy to build different variants, and ensures that our package is always spack-installable.

\* those images are currently built by hand, but i'm thinking to build them every x days on kubernetes, that's easy to setup and doesn't require pr's to SIRIUS.


haampie:
I've added a cron job that builds those two images daily on gitlab, s.t. they can use CSCS's kubernetes cluster with 12 cores per node instead of 1 core (?) with github actions.

- the repo: https://gitlab.com/cscs-ci/electronic-structure/build-spack-images/-/tree/master
- the cronjob: https://gitlab.com/cscs-ci/electronic-structure/build-spack-images/-/pipeline_schedules
- you can also trigger the builds by hand: https://gitlab.com/cscs-ci/electronic-structure/build-spack-images/-/pipelines

haampie:
bors try

haampie:
bors try

PR #517, opened 2020-08-03
Sync develop and master

haampie:


toxa81:
Please, don't merge! I didn't finish benchmarks of QE/Sirius. It seems that with the new implementation of Davidson solver locking must be enabled by default. Otherwise the time-to-solution increases. On the simple tweaks, orthogonalization should be always enabled, and the all the logic of `orthogonalize/not orthogonalize` auxiliary wave-functions can be removed.

toxa81:
I want to merge it to move forward with version 7.

PR #522, opened 2020-08-06
Changes in the generation of magnetic XC potential

toxa81:
The fillowing is changed:
  * add get_rho_up_dn() function in Density class to generata rho_up and rho_dn from density and magnetisation vector
  * for collinear case, magnetisaion along z itself is used, instead of its absolute value; this should prevent kinks in rhoup, rhodn functions
  * extra loop over theta,phi points was harmless and not necessary -> removed


toxa81:
@simonpintarelli can you check this version with your NLCG? Even without QE. I fixed the way Vxc, Bxc is generated. I'm testing it myself, but your insight will be also helpful.

simonpintarelli:
@toxa81 Thanks I'm getting the same as for my tests when I removed the |.| and adapted the rest accordingly. "Noise" is still there.

What is the reason for https://github.com/electronic-structure/SIRIUS/pull/522/commits/d17f1afb35e92f520558458f2f4f8705461bef9a
`align Bxc antiparallel to magnetisation` ?

toxa81:
>What is the reason for d17f1af

It's because magnetic moment of electron is anti-parallel to its spin. And **mag(r)** is actually a spin density.<psi|S|psi>

https://en.wikipedia.org/wiki/Magnetic_moment#Magnetic_moment_of_an_electron

toxa81:
I did a test of FM Ni (GGA) with QE/sirius (`develop` and this branches). With this branch I was able to get a magnetic solution exactly like QE. With `develop` branch it collapses to non-magnetic. Like in the verification test07. So, I declare this an important fix.

toxa81:
@simonpintarelli  @mtaillefumier 
There is a miniapp `test_pppw_xc` now. You can try it like this `/test_pppw_xc --num_mag_dims=0 --xc_name=XC_LDA_C_PW`. There are surprises:

for XC_LDA_X (non-magnetic)
best result :     0.000002014807 for epsilon     0.000001000000
for XC_LDA_X (magnetic)
best result :     0.000000615230 for epsilon     0.000010000000
for XC_LDA_C_PZ (non-magnetic)
best result :     0.000000015949 for epsilon     0.000001000000
for XC_LDA_C_PZ (magnetic)
best result :     0.009890285847 for epsilon     0.100000000000

and so on..

XC_LDA_C_RPA is OK
XC_LDA_C_XALPHA is OK
XC_LDA_X_2D is OK

but most of them will fail (numerical derivative doesn't match the analytical expression).

mtaillefumier:
Thanks. Yes it is a bit surprizing. We must keep in mind that numerical derivatives also have issues (evaluation is sometimes hard). How should we proceed. I can play with some of them in mathematica but not all of them. 
Do we try to implement QE version of them ?

toxa81:
> Thanks. Yes it is a bit surprizing. We must keep in mind that numerical derivatives also have issues (evaluation is sometimes hard). How should we proceed. I can play with some of them in mathematica but not all of them.
> Do we try to implement QE version of them ?

We need to find a way to benchmark XC functionals. Take for example XC_LDA_C_PZ (magnetic). On the simple test of functional derivative it fails. But in the NLCG run it works correctly (right, @simonpintarelli ?). I'm deriving magnetic test in Mathematica. Once this is done there are two options: 1) code the same test in C and use true functionals from libXC or 2) import libXC functionals into Mathematica

simonpintarelli:
@toxa81 Yes, LDA is working fine in the magnetic case.

PR #529, opened 2020-08-18
Pass callback to compute Vloc radial integrals

toxa81:
[WIP] Make periodic function (Vloc in our case) from a list of radial integrals
computed for all lengths of G-shells. Later QE will compute the radial integrals
itself.


toxa81:
Hi @haampie ! What's wrong with elpa mpi builds?

haampie:
Dockerhub being flaky I think, rerun them just now

PR #531, opened 2020-08-19
Fix two more places where elpa set gpu was called

haampie:
And make the config of elpa the same in all solvers


haampie:
@toxa81: not sure if you have had this before, but for me elpa CPU hangs in `tridiag_band` (one process hangs in mpi_wait it seems), it's hard to make a small reproducible example, since it happens after many iterations

toxa81:
@haampie I saw this two. That's why ELPA is not our default parallel eigen-solver

haampie:
Did you submit a ticket / email? I can make the problem go away when both SIRIUS and elpa have openmp disabled

toxa81:
I couldn't come up with a standalone reproducer :/

PR #532, opened 2020-08-19
Remove explicit linking with openmp

haampie:
It already appears conditionally when the OpenMP target is set a few lines below that


toxa81:
jenkins can be ignored here

toxa81:
This should rebased to the latest `develop` branch

PR #534, opened 2020-08-20
Use AND generator expression for HAVE_HIP

haampie:
`-DHAVE_HIP` is currently always set, I believe this is the correct way to using AND in generator expressions


AdhocMan (approved):
LGTM

PR #537, opened 2020-08-21
Fix for the large number of MPI ranks

toxa81:
Debug mode was (correctly) failing on assertion when a null-pointer of the local FFT grid with zero slab-size was accessed.
This is fixed now.

### DIFF
--- a/src/SDDK/memory.hpp
+++ b/src/SDDK/memory.hpp
@@ -657,18 +657,17 @@ void memory_pool_deleter::memory_pool_deleter_impl::free(void* ptr__)
 #ifdef NDEBUG
 #define mdarray_assert(condition__)
 #else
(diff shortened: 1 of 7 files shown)


toxa81:
@haampie Hi Harmen! Will the parallel tests also run? I changed the logic of the `run_tests_*` scripts in ./verification folder. Now external agent has to execute `srun .... ./run_tests....x', i.e. scripts themselves don't know how they are submitted in parallel.

haampie:
Ah, I can push a commit to this branch that adds k-point parallel tests. It's basically just adding an entry to this list: https://github.com/electronic-structure/SIRIUS/blob/develop/ci/.gitlab-ci.yml#L37-L115

haampie:
bors try

PR #540, opened 2020-08-24
nlcg / cmake 

simonpintarelli:
- cmake: avoid linking against GSL cblas, ~expect openblas if MKL is not used~
- pending changes in python implementation / update nlcg api


simonpintarelli:
bors try

PR #541, opened 2020-08-25
Simplify readme instructions

haampie:
I've updated the installation instructions to prefer spack installs by default.

- Removes the Dockerfile bits: I don't really see the point of the docker image with deps installed via apt. It's quite verbose and installs generic binaries, also we're not testing if the Dockerfile still works. The Dockerfile with spack has many anti-patterns... so removed that as well.
- Adds one-liner installs with spack instead.
- Removes `prerequisite.py`, it's outdated, we have spack which is also python based.
- Adds instructions on how to start developing / debugging a certain variant of SIRIUS using spack
- In the CMake instructions, remove all the `X_Root` things and recommend to use `CMAKE_PREFIX_PATH=path1;path2;path3` instead


haampie:
Ok, so I have to update the github action to use the spack docker image too instead of prerequisite.py

toxa81:
What if you do development and need to build SIRIUS with CMake. You need all the packages to be available (somehow) on your platform, but then pass all the info to CMake. This was the whole purpose of this excersise with Dockerfile examples.

README.md
```diff
@@ -45,211 +42,119 @@ The following functionality is currently implemented in SIRIUS:
...
-RUN echo "spack load --dependencies py-mpi4py %gcc@9.2.0" >> /etc/profile.d/spack.sh
-RUN echo "spack load --dependencies hdf5 %gcc@9.2.0" >> /etc/profile.d/spack.sh
-RUN echo "spack load --dependencies gsl %gcc@9.2.0" >> /etc/profile.d/spack.sh
-RUN echo "spack load --dependencies spfft %gcc@9.2.0" >> /etc/profile.d/spack.sh
+```bash
+$ mkdir SIRIUS/build && cd SIRIUS/build
+$ spack install --only=dependencies sirius@develop build_type=Debug +cuda
+$ spack build-env sirius@develop build_type=Debug +cuda -- cmake ..
+$ spack build-env sirius@develop build_type=Debug +cuda -- make -j$(nproc)
+```
```
haampie, line 104: @toxa81, this is where you can directly call cmake / make as you wish.

haampie:
wow, the checkout action requires git 2.18 to be installed in the container, ubuntu 18.04 ships with 2.17, that's quite a requirement :/

simonpintarelli (approved):
lgtm

haampie:
@simonpintarelli there's still a reference to `prerequisite.py` in the section about QE with SIRIUS:

https://github.com/electronic-structure/SIRIUS/blob/dfe879943ab7812fba59028e023a8307880570c9/README.md#quantum-espresso

Your `q-e-sirius` package landed in spack right? What should be the spec? `spack install q-e-sirius ^... ^...`?

simonpintarelli:
@haampie Sorry, I've missed that one. Yes, `spack install q-e-sirius` should do already, and then one can use `spack install q-e-sirius ^sirius+cuda+openmp ...` for example.

haampie:
I've removed the section about cuSOLVER vs elpa for now, because I really can't get elpa to work, and it would otherwise suggest you should install elpa when installing QE SIRIUS.

PR #542, opened 2020-08-26
Use std::filesystem or boost::filesystem for finding species files relative to input json file

haampie:
Maybe this is a bit of overkill, and I'm happy to change this PR, but using `std::filesystem` / `std::experimental::fileystem` / `boost::filesystem` makes it a bit easier to have `sirius.scf --input=path/to/sirius.json` resolve paths to species files relative to the json file instead of relative to the current working directory; and it would still support absolute paths too.

I found that GCC 5.x is currently required to build SpFFT, and [GCC 5.4 already supports `std::experimental::filesystem`](https://godbolt.org/z/W37h6K). Also [clang 3.5.1](https://godbolt.org/z/rvcsKv) seems to support it. Otherwise there's a fallback to `boost::filesytem` (which can be installed with `spack install sirius +boost_filesystem` or `apt-get install libboost-filesystem-dev`).

I haven't checked other compilers yet


haampie:
bors try

toxa81:
I would like to hear opinion of @mtaillefumier @AdhocMan @simonpintarelli on using part of boost. In principle, NLCG library is also introducing a hard dependency on Kokkos, so...

mtaillefumier:
boost is pretty much installed everywhere since many packages depend on it. So if you think it helps I do not see any argument against it especially if the functionality is header only. 

Is NLCG becoming a hard dependency as well or not ? Boost is simple to handle but kokkos is not completely straightforward I think.

simonpintarelli:
Doesn't std::filesystem require C++17? Can we move to C++17? Then we do not need boost.

On the other boost is in every package manager, drawback would be that compilation time goes up (e.g. with spack).

NLCG is optional. Also kokkos does not depend on boost.

simonpintarelli:
Ok, nvcc certainly does not support it, but the cuda kernels are compiled separtely and they don't need c++17. For the intel compiler I don't know.

toxa81:
C++17 is fine as long as users can build the code. It has a good support of special functions  (https://en.cppreference.com/w/cpp/numeric/special_functions). My only concern is some conservative environments with old GCC/Intel

haampie:
> Doesn't std::filesystem require C++17?

It does, but std::experimental::filesystem works with -std=c++11 on GCC 5.4 for instance. The `FindFilesystem` CMake module I added in this PR makes the c++17 dependency explicit through `set_property(TARGET std::filesystem APPEND PROPERTY INTERFACE_COMPILE_FEATURES cxx_std_17)`, but it still seems to work with GCC 5.4 too.

> drawback would be that compilation time goes up (e.g. with spack).

It's not too bad on my laptop if you just install what you need:

```
$ time spack install boost cxxstd=14 +filesystem ~atomic ~chrono ~date_time ~exception ~graph ~iostreams ~locale ~log ~math ~program_options ~random ~regex ~serialization ~signals ~syste
m ~test ~thread ~timer ~wave
==> Installing boost
==> No binary for boost found: installing from source
==> Fetching https://dl.bintray.com/boostorg/release/1.74.0/source/boost_1_74_0.tar.bz2
################################################################################################################################################################################################################################## 100.0%
==> boost: Executing phase: 'install'
[+] /home/hstoppels/spack/opt/spack/linux-ubuntu18.04-skylake/gcc-8.4.0/boost-1.74.0-6cntmawsagsreu75griamkawfisy4ng6

real    0m56.547s
user    1m27.193s
sys     0m11.573s
```

Maybe another concern is that CP2K seems to be proud of their compiler support: https://www.cp2k.org/dev:compiler_support which goes back all the way to GCC 4.9.2. So that's why I included boost as an optional dependency

mtaillefumier:
I understand the concern. I forgot about nvcc actually. If we want a larger compatibility then I think boost cover more compilers than c++17. We can maybe use std::system when the compiler supports it and boost if it does not. In particular if the syntax is identical or very similar between boost and c++17.

haampie:
> I understand the concern. I forgot about nvcc actually. If we want a larger compatibility then I think boost cover more compilers than c++17. We can maybe use std::system when the compiler supports it and boost if it does not. In particular if the syntax is identical or very similar between boost and c++17.

That's what this PR does right now; first it tries to find std::filesystem, then std::experimental::filesystem. If none are found, then it requires Boost's filesystem component.

Should we support `std::experimental::filesystem` though? It would make life easy with GCC 7, which is still the default compiler on ubuntu 18.04. I don't really know what the 'experimental' means here and whether it is safe to use.

mtaillefumier:
You also have to consider the fact that fortran has not changed since 2003 (no real major changes compared to c++) so supporting compilers that are 7 (sorry 6) years old is not a big issue. 

We can not say the same with c++ so it is a question of trade-off. I think gcc 7 should be a reasonable choice as default compiler. It had time to settle and is supported by nvcc. The problem comes when we use very recent compilers and cuda. 

the word "experimental" might mean many things, either that the code is unstable or that the functionality is part of the next standard but not officially published yet.

AdhocMan:
It looks like the functionality required is mainly splitting / merging of paths. Since we are supporting only Linux / macOS, it shouldn't be too difficult to do this "by hand". It would be less clean though, so maybe just as a fall back option.

mtaillefumier:
> 
> That's what this PR does right now; first it tries to find std::filesystem, then std::experimental::filesystem. If none are found, then it requires Boost's filesystem component.
Sorry I missed part the reply. Read it a bit too quickly. If you already do this then I would not be too concerned about the compiler.

mtaillefumier:
> It looks like the functionality required is mainly splitting / merging of paths. Since we are supporting only Linux / macOS, it shouldn't be too difficult to do this "by hand". It would be less clean though, so maybe just as a fall back option.

Can we avoid reinventing the wheel all the time ;).

haampie:
Yeah, obviously we can implement it ourselves too, but writing e.g.

```
auto path = path_to_file.parent_path() / other_path;
```
is already non-trivial... extracting the directory from a path to a file, and then checking whether `other_path` is absolute or not: if it's absolute, disregard the left-hand-side of `/` and just use the right-hand-side; if it's relative, join the paths with the correct directory separator. If someone has already implemented a non-buggy version of that I'd rather use it.

simonpintarelli:
Is it important to have this feature? When running calculation using aiida, files will always be in the same directory. Also the species files can be copied as a string into `sirius.json`, for example using `jq`.

haampie:
It's not a high-priority feature for sure, but being able to write

```
docker run -v `pwd`:/test_sources sirius_image sirius.scf --input=/test_sources
```

instead of

```
docker run -v `pwd`:/test_sources sirius_image /bin/bash -c 'cd /test_sources && sirius.scf'
```
would be better 🤷‍♂️

haampie:
bors try

toxa81 (approved):
lgtm

haampie:
Hm, spack is not picking up the new variants in the `package.py` from our repo, I guess it will never update a package.py in an existing spack environment.

haampie:
bors try

haampie:
bors merge

PR #544, opened 2020-08-27
Replace GPU_MODEL with CUDA_ARCH

haampie:
- support multiple CUDA archs
- map Spack's `cuda_arch` to `CUDA_ARCH` for SIRIUS 7.x instead of fiddling with `CMAKE_CUDA_FLAGS` in package.py
- Use `CUDA_ARCHITECTURES` when CMake 3.18 is used instead of appending flags to `CMAKE_CUDA_FLAGS` in CMake.


AdhocMan (approved):
LGTM

PR #547, opened 2020-08-28
Add a basic clang 6 build as well

haampie:
Using MPICH without Fortran (cause I can't get it to build with a mixed clang/gfortran compiler setup).

Also moves some filesystem remnants from src/ to apps/


haampie:
Hm, when not building Fortran bindings, I guess it's not necessary to require `MPI_Fortran` in CMake? Let me see if that's patchable.

haampie:
bors merge

PR #549, opened 2020-08-31
Use ctest for verification tests too

haampie:
the idea is to be able to run `ctest` to run different setups for different verifications tests.

this currently would provide more or less the same functionality as reframe, since `sirius.scf` seems to be running the same actual tests already when `--test_against=...` is set.


haampie:
Unfortunately I have to merge into develop to start building new base docker images

PR #551, opened 2020-08-31
in the git workflow checkout 2 steps in history

toxa81:


toxa81:
It works, but I don't want to merge this PR now. Let is sit here for a while, until we update more urgent stuff.

PR #554, opened 2020-09-08
cleanup energy functions

toxa81:
A bit more consistent way to compute various total energy contributions


toxa81:
Hi @simonpintarelli ! I think I might have broken the NLCG interface. Now I remember that the energy functions were moved out of `DFT_ground_state class` deliberatlely. Can you add the build of NLCGLIB to our github actions such that SIRIUS is compiled with it. So that I know if my changes are harmful or not.

PR #557, opened 2020-09-16
Integration of SPLA - replaced wf_inner / wf_trans

AdhocMan:


haampie:
bors try

haampie:
bors try

haampie:
Ah, i guess the current build failure is legit: https://gitlab.com/cscs-ci/electronic-structure/SIRIUS/-/jobs/741257907#L2164

haampie:
bors try

haampie:
bors try

haampie:
bors try

AdhocMan:
bors try

AdhocMan:
bors try

AdhocMan:
bors try

toxa81 (commented):
Let me just summarize the top view on this PR.
1. Simulation context gets a new member `spla::Context(SPLA_PU_HOST)`. This will be used to interact with SPLA library. In case of PU_DEVICE SPLA will be using cuda/rocm blas functions. How about mixed pointers? Does SPLA handle this case?
2. dmatrix class gets a new member `spla::MatrixDistribution spla_distri_`. This is used in the same way as BLACS context. From now on dmatrix class depends on spla header
3.  inner(), orthogonalize() and transform() routines are rewritten to use spla; spla context is passed as parameter

AdhocMan:
Thanks, that summary pretty much covers this PR.
With these changes, spla becomes a mandatory dependency, so I'd say usage of the spla header should not be a problem.
In general, spla can handle any combination of host and device pointers, if GPU is set as processing unit. With the old implementation of inner(), the output was updated on host and device. Looking at how the output is used later on, this appears to be necessary. That's why, when calling spla in inner(), the output pointer is always located on host, and if required, the entire matrix is copied to the device as a last step.

Once this PR is merged, I'd like to look at using spla also for local gemm computations. Compared to cuBLAS-XT, the library also supports ROCm and optional processing on Host for small matrices below a given threshold.

toxa81:
>Looking at how the output is used later on, this appears to be necessary. That's why, when calling spla in inner(), the output pointer is always located on host, and if required, the entire matrix is copied to the device as a last step.

Actually, this is not always true (yes, it's a mess :). One example is when we have cuSolver or magma with gpu pointers following the inner() function. In this case the result of inner() should stay on GPU and used inside eigen-solver library without going to CPU. But I think this can be fixed on SIRIUS/inner() side, not on SPLA which will write to a given output pointer, located on CPU or GPU.

AdhocMan:
The behavior of inner() was a bit inconsistent, since if device memory was given as parameter, the output was updated on host and device if not distributed, or only on host if distributed. That probably made sense overall, but it is not really what one would expect. With these changes now, the matrix is always updated on host and device. Not ideal performance wise, and can certainly be improved, but it's less likely to cause issues.
I think it would be helpful in general, to have some sort of mechanism to indicate which memory type is up-to-date. Then each function could work more independently, but that would be quite a big change. What I could do however, is to add the output memory parameter again, but this time **only** update at the given location. Then some memory copies would have to be added outside of inner(), if required. I'm not entirely sure if that would be a cleaner solution then what we had before, but at least inner() would operate consistently.

toxa81:
>With these changes now, the matrix is always updated on host and device. Not ideal performance wise, and can certainly be improved, but it's less likely to cause issues.
Let's keep it like this and make a mental note to revisit this problem.
>I think it would be helpful in general, to have some sort of mechanism to indicate which memory type is up-to-date. 
Exactly! We can ask @simonpintarelli how Kokkos handles two memories and how they provide a flexibility in picking combinations of input, output data locations.

haampie:
Let me just hijack this pr to see if benchmarking can work.

@siriusbot run

```yaml
spec: sirius@develop ^spla@1.2.0
```

haampie:
Benchmark started with the following settings:

```json
{
    "type": "compare",
    "reference": {
        "spec": "sirius@develop ^spla@1.2.0",
        "repo": "electronic-structure/SIRIUS",
        "sha": "cec8cbfd14a8c81a238f6b8e8163f5955b8ecb39"
    },
    "current": {
        "spec": "sirius@develop",
        "repo": "AdhocMan/SIRIUS",
        "sha": "f7e2b5b1eb33882e714b9cd07d32861395e74c7d"
    },
    "report_to": {
        "repository": "electronic-structure/SIRIUS",
        "issue": 557,
        "type": "pr"
    }
}

```

haampie:
oh... `reference` and `current` are messed up. Let me try again:

@siriusbot run

```yaml
spec: sirius@develop ^spla@1.2.0
```

haampie:
Benchmark started with the following settings:

```json
{
    "type": "compare",
    "reference": {
        "spec": "sirius@develop",
        "repo": "electronic-structure/SIRIUS",
        "sha": "cec8cbfd14a8c81a238f6b8e8163f5955b8ecb39"
    },
    "current": {
        "spec": "sirius@develop ^spla@1.2.0",
        "repo": "AdhocMan/SIRIUS",
        "sha": "f7e2b5b1eb33882e714b9cd07d32861395e74c7d"
    },
    "report_to": {
        "repository": "electronic-structure/SIRIUS",
        "issue": 557,
        "type": "pr"
    }
}

```

electronic-structure:
Benchmark report available at https://github.com/haampie/SIRIUS-bench-reports/tree/master/cec8cbfd14a8c81a238f6b8e8163f5955b8ecb39_vs_f7e2b5b1eb33882e714b9cd07d32861395e74c7d/

toxa81 (approved):

toxa81:
@mtaillefumier @simonpintarelli the hard dependency on SPLA is introduced. Please check if it breaks your way of compilation. With spack it should be straightforwards.

mtaillefumier:
is SPLA using cmake as a build system or is it directly introduced in
sirius code. Either way it should not be a big issue.

haampie:
it's cmake and it's living in its own repository

AdhocMan:
It's located here: https://github.com/eth-cscs/spla
The spack package has already been updated to include it when building the develop branch.

mtaillefumier:
Thanks guys. Very useful to know. Then it should not be an issue.

PR #558, opened 2020-09-16
fix nlcglib spec in ci

simonpintarelli:


haampie:
bors merge

haampie:
bors merge

simonpintarelli:
bors merge

simonpintarelli:
bors merge

haampie:
i think the cuda arch issue is a package.py issue? also it seems like it's building/using cmake 3.18.x now, for which the cmake script is a bit different

PR #559, opened 2020-09-16
Drop cmake@:3.14, use latest instead

haampie:
nlcglib needs cmake 3.15+, so let's just use the latest cmake version.

Also add this to the spack environment, s.t. kokkos is cached


haampie:
bors merge

PR #563, opened 2020-09-22
Fix/xc potential

toxa81:
The long standing bug with magnetic GGA (and LSDA as well) is found and fixed. Now each component of XC is computed separately at a small cost of extra manipulations with gradients and Laplacians. Also, SpFFT dependency version is updated to 0.9.13.


toxa81:
bors try

haampie:
i've restarted that one job, seems like spurious daint filesystem error

PR #564, opened 2020-09-25
Fixes for locking and converge_by_energy enabled

haampie:
Fixes a bug where the old eigenvalues were not stored, and does locking
based on the is_converged function now; previously it would take a wrong
number of vectors to lock based on the residual norms of the unconverged
residuals according to the is_converged(..) criterion, which does not
make sense.

Also resets the block_size to num_bands, which makes locking happen more
often in the AUSURF test case.

converge_by_energy is the default again now.


haampie:
bors try

PR #568, opened 2020-09-30
add start/stop pairs to timer

toxa81:
I can then load the pairs of the 1st level nodes, create a flat list of start/stop events and then check the gaps between previous stop and a new start

### DIFF
--- a/apps/timers/find_gaps_in_time.py
+++ b/apps/timers/find_gaps_in_time.py
@@ -1,0 +1,29 @@
+import json
+
+with open('timers.json') as json_file:
+    timers = json.load(json_file)
+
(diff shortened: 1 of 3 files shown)


AdhocMan:
I've slightly changed it, such that only start times are stored / exported, since the end time can easily be calculated with start + duration.

toxa81:
Good! Thanks, @AdhocMan 
It actually helps a lot! In a moment I know that the unmeasured time is somewhere at the end, probably the I/O:

```
(218.283, 'sirius_api::sirius_get_wave_functions', 'stop')
(252.4349, 'sirius::finalize', 'start')
```

PR #570, opened 2020-10-01
Change memory assertion report to stderr to fix CI reporting

AdhocMan:
Added a test assertion, which should fail on one rank in CI. Will be removed before merge.

### DIFF
--- a/src/SDDK/memory.hpp
+++ b/src/SDDK/memory.hpp
@@ -657,18 +657,19 @@ void memory_pool_deleter::memory_pool_deleter_impl::free(void* ptr__)
 #ifdef NDEBUG
 #define mdarray_assert(condition__)
 #else
-#define mdarray_assert(condition__)                                  \
-{                                                                    \
-    if (!(condition__)) {                                            \
-        std::printf("Assertion (%s) failed ", #condition__);         \
-        std::printf("at line %i of file %s\n", __LINE__, __FILE__);  \
-        std::printf("array label: %s\n", label_.c_str());            \
-        for (int i = 0; i < N; i++) {                                \
-            std::printf("dim[%i].size = %li\n", i, dims_[i].size()); \
-        }                                                            \
-        raise(SIGABRT);                                              \
-    }                                                                \
-}
+#define mdarray_assert(condition__)                                                                                    \
+    {                                                                                                                  \
+        if (!(condition__)) {                                                                                          \
+            std::fprintf(stderr, "Assertion (%s) failed ", #condition__);                                              \
+            std::fprintf(stderr, "at line %i of file %s\n", __LINE__, __FILE__);                                       \
+            std::fprintf(stderr, "array label: %s\n", label_.c_str());                                                 \
+            for (int i = 0; i < N; i++) {                                                                              \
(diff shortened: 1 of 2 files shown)


AdhocMan:
bors try

AdhocMan:
bors try

src/SDDK/memory.hpp
```diff
@@ -657,18 +658,18 @@ void memory_pool_deleter::memory_pool_deleter_impl::free(void* ptr__)
...
+    {                                                                                                                  \
+        if (!(condition__)) {                                                                                          \
+            std::fprintf(stderr, "Assertion (%s) failed ", #condition__);                                              \
+            std::fprintf(stderr, "at line %i of file %s\n", __LINE__, __FILE__);                                       \
+            std::fprintf(stderr, "array label: %s\n", label_.c_str());                                                 \
+            for (int i = 0; i < N; i++) {                                                                              \
+                std::fprintf(stderr, "dim[%i].size = %li\n", i, dims_[i].size());                                      \
+            }                                                                                                          \
+            MPI_Abort(MPI_COMM_WORLD, 1);                                                                              \
```
toxa81, line 670: I want to object to this change. First, memory.hpp (memory management functions) now depends on MPI header. Why should they? Second, with raise(SIGABRT) I can get a core dump. This is super valuable on the parallel debugs on Daint. With MPI_Abort the only alternative would be to use distributed debugger, probably with the GUI. Otherwise how I would trace the source of assertion?

haampie: Good point about core dumps, they used to be uploaded to Gitlab before we had `ctest` to run tests. Would be neat to bring that feature back to debug CI segfaults :). Also I'm thinking maybe we had `--unbuffered` in slurm before to ensure data would actually print before a segfault, is that correct @toxa81? Or how else can we explain there's no assert statement being printed?

AdhocMan: Yes, some good points. The hope was that `MPI_Abort` solves the deadlock issue, if an assertion is triggered in CI. Turns out, it doesn't actually help. I'll revert that change.

toxa81: yes, ` --unbuffered` tells slurm to write output immediately instead of storing it in some buffer and writing it in chunks

haampie:
bors try

haampie:
bors try

haampie:
bors try-

haampie:
bors try

haampie:
Seems like there's an issue with cmake/ctest swallowing stderr, even when adding an explicit fflush :thinking:

AdhocMan:
Looking at the ctest [documentation](https://cmake.org/cmake/help/latest/manual/ctest.1.html), maybe the flags `--extra-verbose`, `--output-on-failure` or `--stop-on-failure` could be helpful (if we are not already using them).

haampie:
bors try

AdhocMan:
With the assertion messages now showing up for all ranks, I think these changes are ready for merging.
@haampie do you want to do further testing / additional changes?
Otherwise I'll clean up the commits and mark it ready for review.

haampie:
Should be fine for me

AdhocMan:
bors try

haampie (approved):

toxa81 (approved):
Fine!

PR #582, opened 2020-11-05
Refatoring on LDA+U part

toxa81:
Immediate steps: test12 fails with slightly different total energy. I will run base development branch and compare the occupation matrices. I don't think that the averaging of radial integrals w.r.t. averaging or radial functions can lead to such a significant change.


src/hubbard/hubbard_occupancies_derivatives.cpp
```diff
@@ -318,71 +315,93 @@ Hubbard::compute_occupancies_stress_derivatives(K_point&                    kp__
...
+             *
+             * - kp.hubbard_wave_functions() are the (potentially) orthogonalized functions with S applied
+             * - kp.hubbard_atomic_wave_functions() are the non-orthofonal (initial) atomic Hubbard functions with S applied
+             * - both sets are spin-dependent, even if the orbitals have no spin label
+             * - atom_type.indexr_hub() is the new index of Hubbard radial functions
+             * - atom_type.indexb_hub() is the new index of Hubbard orbitals (radial function times a spherical harmonic)
+             * - wavefunctions_strain_deriv() and K_point::generate_atomic_wave_functions() work with the new indices
+             */
+
+            ////if (ctx_.processing_unit() == device_t::GPU) {
```
toxa81, line 341: @mtaillefumier Mathieu! I need a bit of your help from this line down to the end of the function. You can leave the GPU code for later, I can do it myself. 

|phitmp> contains d|phi>/d eps and S operator is not yet applied

toxa81:
@mtaillefumier I've done with my changes for now

toxa81:
@haampie how I can fix the hdf5 problems in build_cuda_11 build?

toxa81:
do we have to fix smth. on our side?

haampie:
No, this happens because the hdf5 package was misconfigured (api=none as default value, but it's not among the list of acceptable options), and it only shows up now because spack has a new concretizer, and it resulted in it just picking a random acceptable instead. So I fixed it by making the sirius depend on `hdf5 api=v18` in our package.py. It has to build hdf5 now, but binaries will be cached later.

toxa81:
I tested GPU band-parallel test20 (the one which fails in CI). It works, so I think we are OK to merge the PR.

PR #587, opened 2020-11-18
Prepend SIRIUS_* to defines

haampie:
q-e-sirius defines `__CUDA` with a certain build flag and we do too, which might clash when linking. It didn't cause problems for me, but it's confusing anyhow to figure out where a define such as `__CUDA` comes from when looking at the build logs, whereas `SIRIUS_*` defines are very clear. I believe SpFFT has a similar naming scheme for defines.


AdhocMan (approved):
I like this change :+1: 
It's definitely better to have project specific macro names, although I think it's unlikely to cause linking issues with QE, since we mostly use target specific compile definitions in CMake, which do not influence other targets, and no symbols are generated with that name. Issues are more likely with external headers, if macro names happen to collide.

simonpintarelli (approved):
I think its mandatory to use a prefix, otherwise conflicts might arise with any library using SIRIUS.

PR #588, opened 2020-11-23
Restore computation of the hubbard contribution to the stress tensor

mtaillefumier:
- added an element to k_point class containing the atomic orbitals before applying S

need a better naming for the original hubbard orbitals. I can add more comments as well.

### DIFF
--- a/src/hubbard/hubbard_occupancies_derivatives.cpp
+++ b/src/hubbard/hubbard_occupancies_derivatives.cpp
@@ -219,9 +219,18 @@ Hubbard::compute_occupancies_stress_derivatives(K_point& kp__, Q_operator& q_op_
 {
     PROFILE("sirius::Hubbard::compute_occupancies_stress_derivatives");
 
-    auto& phi = kp__.hubbard_wave_functions();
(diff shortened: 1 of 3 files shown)


toxa81:
@mtaillefumier how about naming `hubbard_atomic_wave_functions` and `hubbard_atomic_wave_functions_S` for initial atomic |phi> and S|phi> states and `hubbard_wave_functions` the final set of orbitals, participating in the Hubbard potential correction?

toxa81:
alternatively, for the first two: `atomic_wave_functions_hub_' and 'atomic_wave_functions_S_hub_'

mtaillefumier:
Makes a lot of sense actually.

mtaillefumier:
I may go for the second option. It seems less confusing to me. I will make the changes.

toxa81:
I also  like the second option more. It goes well with the notation `indexb_hub`, `inder_hub`

mtaillefumier:
Indeed and it also tells that they are the original wfc. The other notation might confuse us saying those wfc are the one used in the computation. I will make the necessary changes and push --force to my branch.

PR #589, opened 2020-11-30
Make RMS^2 in SIRIUS the same as dr2 in QE

haampie:


haampie:
This requires https://github.com/electronic-structure/q-e-sirius/pull/10.

I didn't add the metal energy term and scf energy term here.

haampie:
bors try

haampie:
Hm, test20 seems a bit troublesome. I can make it converge in fewer iterations by making the davidson tolerance depend on the RMS value only (so itsol_tol_scale_[1] == 1.0), but the final total energy is `-165.02437739` with 10 digits converged while the desired value is -165.0263349.

toxa81:
test20 is in transition state. I think, systematization of lda+u occupation matrix is not working there. I propose to ignore the results of test20 at the moment. If other tests are working fine, your code works.

haampie:
@electronic-structure run

```yaml
spec: sirius@develop +scalapack
```

electronic-structure:
<details>
<summary>Benchmark <code>benchmarks/62922ab</code> submitted:</summary>

```json
{
    "id": "705e2db0-a27e-45ea-ba3b-63c95411e2af",
    "reference": {
        "spec": "sirius@develop +scalapack",
        "cmd": [
            "sirius.scf"
        ],
        "repo": "electronic-structure/SIRIUS",
        "sha": "abb5f126a26d45879907ac186cd0aa499a19987b",
        "build": true
    },
    "current": {
        "spec": "sirius@develop +scalapack",
        "cmd": [
            "sirius.scf"
        ],
        "repo": "haampie/SIRIUS",
        "sha": "94a9841b4de3015f2d3ffa1886b79486161d7cba",
        "build": true
    },
    "report_to": {
        "repository": "electronic-structure/SIRIUS",
        "type": "pr",
        "issue": 589
    }
}

```
</details>

To trigger benchmarks again using cached binaries, include the following in your comment:
```yaml
id: 705e2db0-a27e-45ea-ba3b-63c95411e2af
```

toxa81:
@haampie Do I need to have a look at this branch?

haampie:
Yeah, this should go together with https://github.com/electronic-structure/q-e-sirius/pull/10 to make the mixer history size, the rms value, the mixer convergence criterion (only based on dr2=sqrt(rms) in QE) and the convergence tolerance in Davidson the same in SIRIUS as it is in QE.

haampie:
It shouldn't really change anything for "standalone" SIRIUS dft, since the energy difference convergence criterion is still around; from QE it is just set to something large so that it's effectively not used.

src/input.hpp
```diff
@@ -712,7 +712,7 @@ struct Settings_input
         ctx_.iterative_solver_tolerance(tol);
         \endcode
      */
-    std::array<double, 2> itsol_tol_scale_{{0.001, 0.5}};
+    std::array<double, 2> itsol_tol_scale_{{0.1, 0.5}};
```
haampie, line 715: This is compensated for with `/ num_electrons` above

src/mixer/mixer_functions.cpp
```diff
@@ -136,7 +136,7 @@ FunctionProperties<Periodic_function<double>> periodic_function_property_modifie
 {
     auto global_size_func = [](Periodic_function<double> const& x) -> double
     {
-        return x.ctx().unit_cell().omega();
+        return 1.0 / x.ctx().unit_cell().omega();
```
haampie, line 139: and this is where you suggested to take `1.0`, instead of `1.0 / omega`, but I left it as one over omega because it corresponds with what QE does. We can make it 1.0, but then there should be an omega^2 term in `set new tolerance of iterative solver` to ensure QE and SIRIUS still set the same threshold for Davidson.

toxa81: I checked: you are  correct. 
`V_{Ha}({\bf G}) = 4\pi \frac{\rho({\bf G})} {{G^2}}` and so inner product has `\Omega` prefactor

haampie:
Finally I'm thinking the convergence criterion for the case `converge_by_energy{false}` could also be improved; near convergence with well-separated eigenvalues it should hold that |λ - θ| = O( ||Ax - λBx||^2 ), we're using ||Ax - λBx|| < epsilon as a criterion in that case, meaning that we are much more strict there (potentially).

PR #590, opened 2020-12-04
Merge same rms as qe

toxa81:


toxa81:
Almost working, but band parallel tests are slow.. Glitch?

toxa81:
@haampie I'm getting completely different execution times if I run band parallel test by hand. Inside CI container, test03 runs for ~283 sec. If I run it myself (1 node, 4 MPI ranks), I get ~30 sec. Density mixing (sirius::Density::mix) for example, takes 46 sec. vs. 0.4 sec. What can it be?

haampie:
Let me try it in a bit. One difference is it's not using MKL but netlib-scalapack+openblas, but I would be surprised that's a factor 100 slower.

haampie:
Ok, could be that because it uses a generic architecture x86_64 in CI the mixer is incredibly slow. Although a 100x slowdown is a bit hard to believe...

The idea of picking a generic architecture was that the container could be run anywhere, so that when a test doesn't pass on Daint, you could download the container and run it on your own hardware to see what's up. I thought it would be OK, because BLAS / CUDA are fast anyways.

toxa81:
I also think it's OK. Do we compile with `-DRelWithDebInfo` or with `-DRelease`? It has become slow after the upgrade. I'm wondering if there was some change in the Sarus or run-time environment...

haampie:
It's compiled with RelWithDebInfo

So, in my comment I was running on my own desktop (amd zen2 arch) in docker, without sarus, so it's likely unrelated to sarus.

I think we can safely assume AVX instructions can be used, so we could also just set the target to Haswell everywhere...

PR #593, opened 2020-12-14
[WIP] addig smearing functions

toxa81:
I'm starting to add smearing functions to the code. For any kind of smearing, three functions should be defined: delta(x) function for the substitution of the real delta-function, occupancy function, which is `\int_{-\inf}^{x} delta(t)dt` and entropy functions, which is `\int_{-\inf}^{x} delta(t) t dt`. In the current implementation I use lambda-function mechanism to select the smearing functions. The same can be done with the base abstract class and overloading. @AdhocMan @simonpintarelli what do you think, which way is better? May be there exists yet another approach?

### DIFF
--- a/apps/tests/test_davidson.cpp
+++ b/apps/tests/test_davidson.cpp
@@ -40,7 +40,6 @@ void init_wf(K_point* kp__, Wave_functions& phi__, int num_bands__, int num_mag_
 
 void test_davidson(cmd_args const& args__)
 {
-    auto pu        = get_device_t(args__.value<std::string>("device", "CPU"));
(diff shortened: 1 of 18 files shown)


simonpintarelli:
I think it's ok to return a std::function, having the function work an a vector and fill all bands in one call might be more efficient. But most likely won't make a difference, since there are not so many bands.

AdhocMan:
I think that using functions is a better solution, if you do not need keep an internal state. However, the compiler cannot optimize function pointers or objects using type erasure like std::function very well. So if performance is important here, it would be better to pass a vector to the function, instead of each element individually.

simonpintarelli (approved):
lgtm

PR #594, opened 2020-12-16
Add generic cg solver

haampie:
Adds a generic cg solver plus a test.

Introduces a test dependency on eigen, cause I found that much easier to test with.


toxa81:
@haampie , we need to merge this one; can you check why clang is complaining?

haampie:
taking a look now

haampie:
bors try

PR #595, opened 2020-12-17
Add entropy contribution

toxa81:
Entropy contribution is non-zero for the metallic systems. All metallic tests have
to be adjusted for the new total energies.


simonpintarelli (approved):
lgtm

mtaillefumier:
good to go. Not much code was changed and everything was adjusted. So can be merged unless something else is broken which I doubt.

PR #599, opened 2021-01-04
Refactor handling of input parameters

toxa81:
Inspired by @mtaillefumier 's `options.json` approach to generate documentation for the CP2K input, I want to move forward and change the way we initialize and set/get simulation context parameters. Eventually, structures like `Mixer_input`, `Iterative_solver_input` etc. should disappear and be replaced by a single JSON dictionary. The dictionary will be created from the JSON schema, **containing the default values**. @mtaillefumier , you actually almost created a true schema (see https://json-schema.org/). As an initial example I'm considering the description of the `mixer` section
```
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "type" : "object",
    "properties": {
        "mixer" :{
            "type" : "object",
            "properties" : {
                "type" : {
                    "type" : "string",
                    "enum" : ["linear", "anderson", "anderson_stable", "broyden2"],
                    "default" : "anderson",
                    "description" : "Type of the mixer."
                },
                "beta" : {
                    "type" : "number",
                    "default" : 0.7,
                    "description": "Mixing parameter"
                },
                "beta0" : {
                    "type" : "number",
                    "default" : 0.15,
                    "description": "Mixing ratio in case of initial linear mixing"
                },
                "linear_mix_rms_tol" : {
                    "type" : "number",
                    "default" : 1e6,
                    "description": "RMS tolerance above which the linear mixing is triggered"
                },
                "max_history" : {
                    "type" : "integer",
                    "default" : 8,
                    "description": "Number of history steps for Broyden-type mixers"
                },
                "beta_scaling_factor" : {
                    "type" : "number",
                     "default" : 1.0,
                    "description": "Scaling factor for mixing parameter"
                },
                "use_hartree" : {
                    "type" : "boolean",
                    "default" : false,
                    "description": "Use Hartree potential in the inner() product for residuals"
                }
            }
        }
    }
}
```
Schema can be extended with the additional fields, for example `cp2k_example`. @mtaillefumier , can you please check that this schema example can be used to generate cp2k documentation, or do they require original `options.json`? In that case I can write a converter from schema to `options.json`.


mtaillefumier:
I would have never guessed we could use this trickery to generate the input keywords. 

cp2k does not require option.json since it use the SIRIUS interface to extract input parameters information. SIRIUS however needs a way to have this internal information (which was the purpose of the options.json file converted to a string) but if you can get the same result with your approach then you do not need to worry about options.json and you can remove it.

toxa81:
How do they generate HTML documentation for the SIRIUS section? Was it crated by you?

mtaillefumier:
it is home cooked cp2k recipy. We have to give a description of the
option when we create the input keywords in cp2k. So what cp2k does is
simple. It calls sirius and ask it how many input sections are there.
Then it goes through all keywords and ask its type, description, name,
and default value which are basically the information included in the
options.json file. CP2K has *no* knowledge of the option.json file at
all because SIRIUS provides all these information.

toxa81:
@teonnik @AdhocMan @haampie 
Guys! I'm observing a strange behavior of cmake (3.19.2). I have a valid json file `input_schema.json` which doesn't contain `\;` sequence of characters. Then it is read by cmake and used to configure a json raw string header
```
# generate schema
file(STRINGS "${PROJECT_SOURCE_DIR}/src/context/input_schema.json"
    SIRIUS_INPUT_SCHEMA NEWLINE_CONSUME)
configure_file("${PROJECT_SOURCE_DIR}/src/context/input_schema.hpp.in"
               "${PROJECT_BINARY_DIR}/src/context/input_schema.hpp"
               @ONLY)
```
After that, the generated `input_schema.hpp` header file contains a json dictionary, BUT, with `\;` sequence, which is incorrect json, and then nlohmann crashes. I tend to believe this is a bug. What do you think?

toxa81:
Update: if I print `message(STATUS ${SIRIUS_INPUT_SCHEMA})` it looks like a valid json.

toxa81:
Hi @haampie ! How I can update spfft to version 1.0.1 for the Daint builds?

haampie:
It should be available automatically in 1 or 2 days after the PR to spack is merged, but to get it right now you can locally add the package.py of spfft with the 1.0.1 version.

toxa81:
Hi @haampie ! All builds pass, why it does not execute the last pipeline on Daint?

haampie:
Maybe it's better how it was before, cause now actions are triggered twice again :/. You can't run the `trigger_gitlab_pipeline` action on a PR because it uses a secret (key), so it has to run on all 'push' events, but that triggers build/test actions twice if you open a PR from the repo itself.

toxa81:
> Maybe it's better how it was before, cause now actions are triggered twice again :/. You can't run the `trigger_gitlab_pipeline` action on a PR because it uses a secret (key), so it has to run on all 'push' events, but that triggers build/test actions twice if you open a PR from the repo itself.

It likes too much the PR which is under development (builds all actions twice :) May be issue a manual command to execute the Daint pipeline? We can discuss it today at the group meeting.

haampie:
that still exists it think, it's

bors try

toxa81:
@simonpintarelli Now the access to nlcg parameters is through ctx.cfg().nlcg() object, for example ctx.cfg().nlcg().maxiter() or ctx.cfg().nlcg().restart(). I removed `smearing` and `processing_unit` items - they should be available through the `paremeters` and `control` sections. Can you check that this is OK for NLCG code?

toxa81:
bors try

toxa81:
First big step of refactoring input parameters is over. All parameters now should be defined in the JSON schema file `input_schema.json`. This schema is used to compose the dictionary of default parameters (see `compose_default_json()` in `context/simulation_parameters.cpp`) which are then overwritten by the user (if needed). With the exception of Hubbard_input section, the rest of parameters is now provided through the JSON dictionary which is saved in the output file. The parser for the JSON configuration dictionary (`context/config.hpp`) is auto-generated from the JSON schema using `apps/utils/gen_input_struct.py` script. The `options.json` file which is used to generate CP2K documentation for the input is not touched in this PR.

toxa81:
bors try

toxa81:
@mtaillefumier @simonpintarelli When you have time, can you review this PR?

mtaillefumier (approved):
I went through twice and could not find anything to object ;). This will simplify our life when we add new things in SIRIUS. So unless Simon finds something i think it is fine to merge.

simonpintarelli (approved):
lgtm

PR #600, opened 2021-01-08
Find HIP libraries through generated config file

AdhocMan:
As suggested by @haampie , custom CMake modules are replaced with config files provided by ROCm. In the past, they were not compatible with other compilers such as GCC, but this is no longer the case, so this is a cleaner solution.


AdhocMan:
Looks like the HIP config files in our CI are broken. Will have check if that's a general issue with ROCm 4.0.

haampie:
This might be the first CI run with ROCm 4.0, so it could be unrelated tot he PR...

AdhocMan:
I've tested it with a fresh install from their Ubuntu repository and it worked fine. So it seems to be related to how spack installs ROCm.
But I also think the way they determine the path for the linker, is not good practice. They look for the header file `stddef.h` [here](https://github.com/ROCm-Developer-Tools/HIP/blob/main/hip-config.cmake.in#L84) and without checking whether that file actually exists, use that variable with an appended relative path for locating libraries.

haampie:
Yeah, this is the questionable cmake practices I was talking about...

The whole branch here is questionable: https://github.com/ROCm-Developer-Tools/HIP/blob/main/hip-config.cmake.in#L67-L90. `${ROCM_PATH}/llvm` is definitely not an existing path to llvm-amdgpu when installed with spack. 

In the end they do a lot of work to add these targets:

```
  set_property(TARGET hip::device APPEND PROPERTY
    INTERFACE_INCLUDE_DIRECTORIES "${HIP_CLANG_INCLUDE_PATH}/.."
  )

  set_property(TARGET hip::device APPEND PROPERTY
    INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${HIP_CLANG_INCLUDE_PATH}/.."
  )

  # Add support for __fp16 and _Float16, explicitly link with compiler-rt
  set_property(TARGET hip::host APPEND PROPERTY
    INTERFACE_LINK_LIBRARIES -L${HIP_CLANG_INCLUDE_PATH}/../lib/linux -lclang_rt.builtins-x86_64
  )
  set_property(TARGET hip::device APPEND PROPERTY
    INTERFACE_LINK_LIBRARIES -L${HIP_CLANG_INCLUDE_PATH}/../lib/linux -lclang_rt.builtins-x86_64
  )
```
I'm pretty sure it's possible to generate a cmake config file which defines the prefix of llvm / clang as an absolute path, so that you don't have to seach for it during find_package(hip). I'll take a look tomorrow

haampie:
I've narrowed the issue down to the following:

```
/usr/bin/g++ \
-march=x86-64 \
-mtune=generic \
...
-o libsirius.so \
... many object files and shared libs ...
/opt/spack/opt/spack/linux-ubuntu20.04-x86_64/gcc-9.3.0/hip-4.0.0-yti4dvpkjsoqc3ce6yetorcmymhvfzlh/lib/libamdhip64.so.4.0.21021-
...
```

Seems like gcc does not like the final dash `-` in the shared library, even though it is a file:

```
file /opt/spack/opt/spack/linux-ubuntu20.04-x86_64/gcc-9.3.0/hip-4.0.0-yti4dvpkjsoqc3ce6yetorcmymhvfzlh/lib/libamdhip64.so.4.0.21021-
/opt/spack/opt/spack/linux-ubuntu20.04-x86_64/gcc-9.3.0/hip-4.0.0-yti4dvpkjsoqc3ce6yetorcmymhvfzlh/lib/libamdhip64.so.4.0.21021-: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=ad24bca9e306e8e10048ff3aa7b16f0867cdddac, stripped
```

Linking works fine when I use the symlink `libamdhip64.so`. I'll check what the suffix is supposed to be.

haampie:
@AdhocMan, the current situation is not great, but it works after specifying `HIP_CXX_COMPILER`. Hopefully this is just temporary as someone in the hip-team suggested they'd drop linking to `clang_rt.builtins-x86_64` for `hip::host`.

And the other compilation error is gone as a result of https://github.com/spack/spack/pull/20780.

PR #605, opened 2021-02-11
SCF correction to forces

toxa81:
If we have SCF correction to total energy, we also need it for the forces.


toxa81:
bors try

mtaillefumier (approved):
looks alright to me.

toxa81:
bors try

PR #606, opened 2021-02-11
Emit warnings for unrecognized config params

haampie:
This one bit me:

`std_evp_solver_type` changed to `std_evp_solver_name` apparently, making benchmarks switch to scalapack instead of cusolver

### DIFF
--- a/src/context/simulation_parameters.cpp
+++ b/src/context/simulation_parameters.cpp
@@ -29,6 +29,9 @@
 #include "context/runtime_options_json.hpp"
 #include "context/input_schema.hpp"
 
+#include <unordered_set>
+#include <iterator>
+
 namespace sirius {
 
 /// Compose JSON dictionary with default parameters based on input schema.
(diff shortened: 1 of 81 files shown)


haampie:
@toxa81 or should I revert `*_evp_solver_type -> *evp_solver_name ` because it breaks the input spec? Are there more breaking changes since versions 6.x?

Asking because benchmarks vs old versions are easiest if we have just one sirius.json input file.

validation/SiO2-ANA/sirius.json
```diff
@@ -1,7 +1,7 @@
 {
     "control" : {
         "processing_unit" : "cpu",
-        "std_evp_solver_type" : "lapack",
+        "std_evp_solver_name" : "lapack",
         "en_evp_solver_type" : "lapack",
```
haampie, line 5: there are many typo's in `validation/` :(

src/context/simulation_parameters.cpp
```diff
@@ -79,6 +95,14 @@ void compose_json(nlohmann::json const& schema__, nlohmann::json const& in__, nl
             }
         }
     }
+
+    // Emit warnings about keys that were set but unused.
+    if (!visited.empty()) {
+        std::stringstream ss;
+        ss << "The following configuration parameters were not recognized and ignored: ";
```
haampie, line 102: This is the important part of the pr ^^

haampie:
bors try

haampie:
Seems like the following entires are not part of the json schema:

```
kpoints_rel
kpoints_path
title
```
should they be?

toxa81:
Proper verification against JSON schema would be in the future. For now, simplest would be to have both `std_evp_solver_type` and `std_evp_solver_name` (and the same for gen_evp) in the json files of the benchmarks.

toxa81:
> Seems like the following entires are not part of the json schema:
> 
> ```
> kpoints_rel
> kpoints_path
> title
> ```
> 
> should they be?

They are not used by the groun state. And the json schema is flexible: extra keys are allowed by default

haampie:
Ok, the hubbards params are brought back again

haampie:
bors try

haampie:
> Proper verification against JSON schema would be in the future. For now, simplest would be to have both std_evp_solver_type and std_evp_solver_name (and the same for gen_evp) in the json files of the benchmarks.

I'd say it doesn't hurt to keep the warnings around, cause it's an easy way to detect typo's etc

haampie:
bors try

toxa81:
LGTM

PR #610, opened 2021-02-25
Updates to the code

toxa81:
- band occupancies are now searched using bisection method and in parallel
- allgather interface is unified: count and offset are provided in the same order
- [WIP] documentation to the occupancy matrix


toxa81:
bors try

toxa81:
bors try

toxa81:
This is strange. I can run the tests on Daint manually and all of them pass.

haampie:
Hm, it seems to be reproducible; fails also after restarting those band parallel tests by hand, just on test 20

toxa81:
bors try

toxa81:
test20 converges to a different total energy, but I think the test itself is not robust enough. So it's ok to merge

toxa81:
bors try

PR #612, opened 2021-03-14
Performance fixes

toxa81:
Fix performance is several places:
* radial integrals of density, core density and local potential are parallelized over all MPI ranks
* access to band energies and occupancies in `K_point` class is improved
* search for Fermi energy is parallelized
* band sums of eigen-values and entropy contribution are parallelized
* test20 is currently removed from the test set
* some minor timers are removed
* the arguments of `Communicator::allgather()` wrappers are consistent
* a fix in GGA stress tensor


toxa81:
bors try

mtaillefumier (approved):
Everything sounds good. mostly minor changes

PR #613, opened 2021-03-15
Deps maintenance

haampie:
- Fix the cuda versions (maybe we should drop cuda as externals to automatically get new patch release versions?)
- Seems like spack doesn't pick up the cmake version so let's stick to a new one.
- No need to specify the hdf5 api version anymore, defaults to 'none'


toxa81:
>Fix the cuda versions (maybe we should drop cuda as externals to automatically get new patch release versions?)
Are all minor versions of cuda-10 compatible with Daint. If yes, I agree, we should build for major version (10).

PR #615, opened 2021-03-16
use smart pointers

toxa81:
Use std::shared_ptr in the lists of atoms, atom types and atom symmetry classes. This allows to add atom type and atoms of that type at the same time. Before, all atom types had to be added first.

### DIFF
--- a/src/unit_cell/atom.hpp
+++ b/src/unit_cell/atom.hpp
@@ -40,7 +40,7 @@ class Atom
     Atom_type const& type_;
 
     /// Symmetry class of the given atom.
-    Atom_symmetry_class* symmetry_class_{nullptr};
+    std::shared_ptr<Atom_symmetry_class> symmetry_class_;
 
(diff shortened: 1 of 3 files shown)


toxa81:
bors try

src/unit_cell/atom.hpp
```diff
@@ -374,7 +374,7 @@ class Atom
     }
 
     /// Set symmetry class of the atom.
-    inline void set_symmetry_class(Atom_symmetry_class* symmetry_class__)
+    inline void set_symmetry_class(std::shared_ptr<Atom_symmetry_class> symmetry_class__)
     {
         symmetry_class_ = symmetry_class__;
```
AdhocMan, line 379: You could use std::move here: `symmetry_class_ = std::move(symmetry_class__);`

AdhocMan (approved):
Looks good to me. For small number of elements, this is a good solution for shared ownership.

PR #616, opened 2021-03-16
merge develop -> master

toxa81:


haampie:
let's see if we have similar slurm issues as DLAF

bors try

PR #620, opened 2021-04-14
Context the hubbard section to the new input scheme

mtaillefumier:
Really nothing ground breaking here. The hubbard input info are using the new input scheme.

### DIFF
--- a/src/context/input.cpp
+++ b/src/context/input.cpp
@@ -27,22 +27,19 @@
 
 namespace sirius {
 
-void Hubbard_input::read(json const& parser)
+void
+Hubbard_input::read(json const& parser)
 {
     if (!parser.count("hubbard")) {
         return;
     }
-
     auto section = parser["hubbard"];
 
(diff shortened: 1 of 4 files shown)


toxa81:
Next step is actually to switch to config.hpp parser in parsing the Hubbard input and get rid of input.cpp. If something is updated in the JSON schema, the config.hpp should be updated by `python ../../apps/utils/gen_input_struct.py > config.hpp`

mtaillefumier:
Ok. Do not merge this then. I want to recheck everything and take into account what you are indicating.

mtaillefumier:
I kept the json structure intact but add to modify the input.cpp to fit the scheme.

toxa81:
Hi @mtaillefumier ! Do we have a use case for `projection_method=file`. If not, I propose to deprecate it until the real need to read Hubbard WFs from the file emerges.

mtaillefumier:
go ahead nuke it. I do not think even QE uses this option.

PR #622, opened 2021-04-29
Tweaks to API generation

toxa81:
Allow `pass by value` for the Fortran API.
This is now possible `call sirius_create_context(MPI_COMM_WORLD, handler)`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # We could be fine with 3.12, but 3.14 has ctest --show-only=json-v1 used in CI.
 cmake_minimum_required(VERSION 3.14)
 
-project(SIRIUS VERSION 7.2.1)
+project(SIRIUS VERSION 7.2.2)
 
 # user variables
 set(CREATE_PYTHON_MODULE    OFF CACHE BOOL "create sirius Python module")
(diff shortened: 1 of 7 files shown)


toxa81:
bors try

toxa81:
Do you know why CMake doesn't pick it? It knows that it is a Fortran/C++ project. It has discovered MPI library and now it needs to link with fortran part

haampie:
So we're adding the Fortran sources to the sirius cmake target, maybe a `$<$<BOOL:${CREATE_FORTRAN_BINDINGS}>:MPI::MPI_FORTRAN>` should be added in src/CMakeLists.txt?

toxa81:
> So we're adding the Fortran sources to the sirius cmake target, maybe a `$<$<BOOL:${CREATE_FORTRAN_BINDINGS}>:MPI::MPI_FORTRAN>` should be added in src/CMakeLists.txt? There's only a MPI_CXX there right now

let's consult with @AdhocMan

simonpintarelli:
I think also, because of the `use mpi`, we need to link with mpi fortran as Harmen said.

AdhocMan:
> I think also, because of the `use mpi`, we need to link with mpi fortran as Harmen said.

I agree, that the error occurs because of calling mpi functions in the fortran test executables. I'd suggest to only link the fortran executables directly to `MPI::MPI_FORTRAN`, since the sirius library itself doesn't require it.

toxa81:
bors try

PR #624, opened 2021-05-04
Enchance the indexing of radial basis functions (Hubbard case)

toxa81:
- [x] Introduce new classes `angular_momentum`, 'radial_functions_index' and `basis_functions_index` in the `experimental` namespace
- [x] Switch to indexing Hubbard wave-functions in the new way
- [x] Refactor symmetrization of occupancy matrix
- [x] Test new symmetrization
- [x] Add lda+u test for NiO collinear
  - [x] Forces need a fix
- [x] Refactor input of Hubbard parameters to a new schema 
- [x] Refactor occupation matrix (change from a single 4D array to a "by-atom" array. Some will be empty because of no Hubbard correction)
- [ ] Add support of multi-orbital case. For example, correct 3d and 4p states on Ni or whatever is available. We are interested in the functionality. 
- [x] Base class for Hubbard occupancy and potential
  - [x] Fix printing of the occupancy information (base class doesn't do - it prints only the matrix elements)
- [x] Have a proper fix when computing the hubbard forces with orthogonalized wavefunctions. ([PRB 102, 235159](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.102.235159))


toxa81:
bors try

toxa81:
bors try

toxa81:
@mtaillefumier I added a new test22 for AFM NiO with LSDA+U. The gap and magnetic moment seem to be reasonable. The atomic forces are wrong. Should be zero, but they are not. May be this is something trivial related to orthogonality of Hubbard WFs. From what I understand the only way to take derivatives w.r.t. lattice strain or atomic displacement is to keep original (non-orthogonal) WFs.

mtaillefumier:
> @mtaillefumier I added a new test22 for AFM NiO with LSDA+U. The gap and magnetic moment seem to be reasonable. The atomic forces are wrong. Should be zero, but they are not. May be this is something trivial related to orthogonality of Hubbard WFs. From what I understand the only way to take derivatives w.r.t. lattice strain or atomic displacement is to keep original (non-orthogonal) WFs. Can you please check it?

You are correct, at least it is implemented that way. The reason behind is that if we take the orthogonal hubbard wavefunctions we mix atomic WFs together. It means that the derivatives of the orthogonal wavefunctions are not simply the derivative or the atomic wave functions that we transform back to the orthogonal set.

|\Psi^k_U> = P_k^i |\psi_i> 

(i is for the atomic wave functions, with Einstein convention). Then when we derive this compared to r_i we get this

\partial_{r_m} |\Psi^k_U> = (\partial_{r_m} P_{k}^{i})   |\psi_i> + P_{k}^{i} (\partial_{r_i} |\psi_i>) 

so we must compute the first term which is far from trivial to do.

toxa81:
@mtaillefumier Then my question is: what would fix the forces in test22?

mtaillefumier:
I can not answer for now. I am reviewing the literature to check that I did not miss something.

mtaillefumier:
By the way, we can use cp2k to answer this question. We calculate the forces numerically and compare to what we expect.

toxa81:
bors try

mtaillefumier:
Made a PR directly to the atomic-basis-index

toxa81:
bors try

toxa81:
something is not allocated on GPUs in K_point::orthogonalize_hubbard_orbital()

toxa81:
bors try

toxa81:
bors try

toxa81:
@mtaillefumier  can you try this branch for cp2k H2 test?

mtaillefumier:
it seems to work. No crash from cp2k.

toxa81:
bors try

toxa81:
@mtaillefumier I propose to postpone the last bullet point: Add support of multi-orbital case. This is not critical for the current action plan.

mtaillefumier:
@toxa81 : No problem with this.

toxa81:
bors try

toxa81:
The PR grew considerably and needs to be merged.

toxa81:
@mtaillefumier do you think we can merge it?

mtaillefumier:
@toxa81 We can merge it to develop. I see no reasons why we should postpone it any further. All tests are green so good to go for me.

PR #627, opened 2021-05-17
Trace radial integral

toxa81:
There is an error in q-point length - radial integrals complain about it. The error is not reproducible in a simple test. The fixes are an attempt to get a trace of what's causing it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # We could be fine with 3.12, but 3.14 has ctest --show-only=json-v1 used in CI.
 cmake_minimum_required(VERSION 3.14)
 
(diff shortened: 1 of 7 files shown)


toxa81:
bors try

simonpintarelli (approved):
lgtm. We could once add specific errors inheriting from `std::excpetion` instead of catching everything.

toxa81:
> lgtm. We could once add specific errors inheriting from `std::excpetion` instead of catching everything.

That's a bigger discussion which I support. How do we handle errors and return a useful info back to Fortran. I was also thinking about the exceptions, but there are just too many of them. So at the moment it is just one std::runtime_error

PR #629, opened 2021-05-19
add missing transpose

simonpintarelli:
conversion from absolute to relative position was wrong when using .unit_cell.atom_coordinate_units=A|au


toxa81 (commented):
The convention is this: r_c = L*r_l, where r_c is vector in Cartesian coordinates and r_l is a vector in lattice coordinates. Then r_l = L^{-1}r_c. There should be no transposition.

toxa81 (changes requested):
just a small change:
/* here lv are copied from the JSON dictionary as three row vectors; however in the
 code the lattice vectors are stored as three column vectors, so transposition is needed here */
auto ilv = inverse(transpose(lv));

PR #631, opened 2021-05-21
fix the calculations of hubbard forces.

mtaillefumier:
- It is a temporary fix since we do not consider the case where hubbard orbitals are reorthogonalized.


toxa81:
@mtaillefumier I think this is too late to change it here. Can you make all the changes into https://github.com/electronic-structure/SIRIUS/pull/624

mtaillefumier:
Sure. Let me change the merging branch.

PR #637, opened 2021-05-26
Hydrogen atoms PP without beta projectors

mtaillefumier:
- SIRIUS will fail if we run H2 molecules with PP that have no beta projectors. The symmetrization procedure will crash in that case, so we do not symmetrize the density matrix (which is not defined in that case). I am not sure this is the right way to fix this corner case and using properly defined PP is probably better. This problem only occurs when the system only contains hydrogen atoms.

### DIFF
--- a/src/density/density.cpp
+++ b/src/density/density.cpp
@@ -63,11 +63,12 @@ Density::Density(Simulation_context& ctx__)
 
     l_by_lm_ = utils::l_by_lm(ctx_.lmax_rho());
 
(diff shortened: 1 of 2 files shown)


toxa81:
bors try

toxa81:
I did it here in a slightly different way
https://github.com/electronic-structure/SIRIUS/pull/624/commits/d1859b5c20bbb4c49c07b80e470a368307522861

The hack `std::max(1, unit_cell_.max_mt_basis_size())` is inconvenient

toxa81:
I think this is now handled by the latest big PR

mtaillefumier:
Ok I can close it then

PR #639, opened 2021-05-30
Add example of Dockerfile

toxa81:
Hi Harmen @haampie !
Can you please have a look and refactor if needed the Dockerfile. I want to have an example of the base container with  sirius installed. It works so far, but if jump inside the container and execute `spack spec sirius@develop` or `spack spec sirius@develop +fortran +scalapack' it starts building some other packages which is weird. Here:
```
root@43002c3f99b0:~# spack spec sirius@develop +fortran +scalapack
Input spec
--------------------------------
sirius@develop+fortran+scalapack

Concretized
--------------------------------
==> Warning: Missing a source id for python@3.8
==> Installing libiconv-1.16-w6zptbclncs3jsnl3i4qtg72fgheiwdf
==> No binary for libiconv-1.16-w6zptbclncs3jsnl3i4qtg72fgheiwdf found: installing from source
==> Using cached archive: /tmp/root/source_cache/_source-cache/archive/e6/e6a1b1b589654277ee790cce3734f07876ac4ccfaecbee8afa0b649cf529cc04.tar.gz
==> No patches needed for libiconv
==> libiconv: Executing phase: 'autoreconf'
==> libiconv: Executing phase: 'configure'
==> Error: Exception occurred in writer daemon!
==> Error: Failed to install libiconv due to KeyboardInterrupt:
Traceback (most recent call last):
  File "/spack/spack_src/lib/spack/llnl/util/tty/log.py", line 754, in _writer_daemon
    rlist, _, _ = _retry(select.select)(istreams, [], [], 1e-1)
  File "/spack/spack_src/lib/spack/llnl/util/tty/log.py", line 837, in wrapped
    return function(*args, **kwargs)
KeyboardInterrupt

==> Error: Keyboard interrupt.
```

### DIFF
--- a/dockerfile/Dockerfile
+++ b/dockerfile/Dockerfile
@@ -1,0 +1,84 @@
+#FROM nvidia/cuda:11.3.0-runtime-ubuntu20.04
+FROM ubuntu:20.04
+
+#==ENV NCCL_VERSION 2.9.6
+#==
+ENV DEBIAN_FRONTEND noninteractive
+#==
+ENV FORCE_UNSAFE_CONFIGURE 1
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  apt-utils python3-distutils python3-apt python3-dev \
+  gcc g++ gfortran python3 git make apt-transport-https ca-certificates \
+  gnupg software-properties-common curl patch pkg-config vim wget xz-utils unzip
+#==
+#==RUN apt-get update && apt-get install -y --no-install-recommends \
+#==    libtinfo5 libncursesw5 \
+#==    gcc g++ gfortran git make \
+#==    cuda-cudart-dev-11-3=11.3.58-1 \
+#==    cuda-command-line-tools-11-3=11.3.0-1 \
+#==    cuda-minimal-build-11-3=11.3.0-1 \
+#==    cuda-libraries-dev-11-3=11.3.0-1 \
+#==    cuda-nvml-dev-11-3=11.3.58-1 \
+#==    libnpp-dev-11-3=11.3.3.44-1 \
+#==    libnccl-dev=2.9.6-1+cuda11.3 \
+#==    libcublas-dev-11-3=11.4.2.10064-1 \
+#==    libcusparse-dev-11-3=11.5.0.58-1 \
+#==    python \
+#==    && rm -rf /var/lib/apt/lists/*
+#==
+#==# apt from auto upgrading the cublas package. See https://gitlab.com/nvidia/container-images/cuda/-/issues/88
+#==RUN apt-mark hold libcublas-dev-11-3 libnccl-dev
+#==ENV LIBRARY_PATH /usr/local/cuda/lib64/stubs
+#==
+# get latest version of spack
+#RUN git clone https://github.com/spack/spack.git /usr/spack
+#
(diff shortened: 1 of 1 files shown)


haampie:
Hi Anton, if you're using spack.x you just need to install compilers in the image/container, the rest (python, make, ...) is already part of spack.x.

> it starts building some other packages which is weird

It'll build all dependencies. Note that we have this base image `stabbles/sirius-cuda-11` in CI exactly because I wanted to avoid having to build dependencies from sources:

```
$ docker run stabbles/sirius-cuda-11 spack find
==> 53 installed packages
-- linux-ubuntu20.04-x86_64 / gcc@9.3.0 -------------------------
berkeley-db@18.1.40  gettext@0.21       libtool@2.4.6     py-botocore@1.20.27       spfft@1.0.3
bison@3.7.6          gsl@2.6            libxc@5.1.3       py-jmespath@0.10.0        spglib@1.16.1
bzip2@1.0.8          hdf5@1.10.7        libxml2@2.9.10    py-python-dateutil@2.8.1  spla@1.4.0
clingo@spack         hwloc@2.4.1        m4@1.4.18         py-s3transfer@0.3.4       sqlite@3.35.5
cmake@3.20.2         libbsd@0.11.3      mpich@3.4.1       py-setuptools@50.3.2      tar@1.34
cuda@11.2.1          libfabric@1.12.1   ncurses@6.2       py-setuptools-scm@4.1.2   util-linux-uuid@2.36.2
diffutils@3.7        libffi@3.3         openblas@0.3.15   py-six@1.15.0             util-macros@1.19.1
expat@2.3.0          libiconv@1.16      openssl@1.1.1k    py-urllib3@1.25.6         xz@5.2.5
fftw@3.3.9           libmd@1.0.3        perl@5.32.1       python@3.8.10             zlib@1.2.11
findutils@4.8.0      libpciaccess@0.16  pkgconf@1.7.4     re2c@1.2.1
gdbm@1.19            libsigsegv@2.13    py-boto3@1.17.27  readline@8.1
```

which could be the base image if you want:

```
FROM stabbles/sirius-cuda-11

RUN spack install sirius@develop target=x86_64 +fortran +scalapack ^mpich
```

But do note it's compiled for a generic architecture `target=x86_64` instead of `target=haswell` or `target=zen2`...

toxa81:
I just added ^intel-mkl as a backend for BLAS/LAPACK

haampie:
I think like this it's fine. You can make `spack env activate .` set environment variables like `LOAD_PATH` and `CPATH` and `LD_LIBRARY_PATH` to make the `-L`, `-I` and `-rpath` flags resp redundant, but I'm not a big fan of environment-variable-based-compilation

toxa81 (commented):
Is there a way to simplify the environment setup? As a user I want to execute `sudo docker run -it --rm sirius` and inside the container already have everything ready to execute `mpifort test.f90 -Iview/include/sirius -Wl,-rpath=view/lib -Lview/lib -lsirius`

haampie:
Let me check

haampie:
This saves an explicit `spack env activate` (actually this is what spack containerize does by default :sweat_smile: ...) and shows how the user can mount their sources into the container, which is the most common use case.

toxa81:
This looks perfect now.

PR #641, opened 2021-06-01
Template develop

manhin321:
Templated eigensolver for lapack which support floating point 32 calculation


toxa81:
bors try

src/SDDK/wf_inner.cpp
```diff
@@ -35,61 +35,77 @@
...
+struct real_type<std::complex<T>> {using type = T;};
+
+namespace {
+//template <typename T>
+//void scale_gamma_wf(spin_range spins, int m, int i0, double alpha, Wave_functions& bra);
 
 // If scalar type, g-vector 0 contribution must be scaled before / after inner product to aboid counting twice
-template <>
-void scale_gamma_wf<double>(spin_range spins, int m, int i0, double alpha, Wave_functions& bra) {
+template <typename F, typename T>
```
toxa81, line 52: Out of my own curiosity: why there are two types: F and T?

src/SDDK/wf_ortho.cpp
```diff
@@ -46,30 +46,31 @@ orthogonalize(::spla::Context& spla_ctx__, memory_t mem__, linalg_t la__, int is
...
             K *= 2;
         }
     }
 
     auto sddk_debug_ptr = utils::get_env<int>("SDDK_DEBUG");
     int sddk_debug      = (sddk_debug_ptr) ? (*sddk_debug_ptr) : 0;
 
-    double ngop{0};
-    if (std::is_same<T, double>::value) {
+    precision_type ngop{8e-9};
```
toxa81, line 60: ngop is the number of floating-point operations (int giga-units) for the matrix-matrix multiplications. For real type it is 2*m*n*k, for complex types it is 8*m*n*k. Basically, it is there to report FLOPS. It can always be a `double` data type.

src/SDDK/wf_trans.hpp
```diff
@@ -27,6 +27,14 @@
 #include <spla/spla.hpp>
 namespace sddk {
 
+// define type traits for a single template implementation of both real and complex matrix
+// general case for real matrix
+template <typename T>
```
toxa81, line 32: again, move to one common place

toxa81 (commented):
Overall it looks good, just a few adjustments are needed.

src/SDDK/wf_inner.cpp
manhin321: It was originally an error to compile that without this definition but now it is fixed using real_type

src/SDDK/dmatrix.hpp
```diff
@@ -265,6 +274,15 @@ class dmatrix : public matrix<T>
             return Communicator::self();
         }
     }
+
+    /* copy the matrix element from another dmatrix with the possibility that they store in different data type which could
+     * which could upgrade or downgrade the precision of the element
+     * Currently only support matrix element copy of host data
+     */
+    template <typename F>
+    void copy_matrix_element(const dmatrix<F>& other_dmatrix){
```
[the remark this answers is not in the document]
manhin321: This is indeed my fault of duplication the code.

src/SDDK/type_definition.hpp
```diff
@@ -0,0 +1,21 @@
+/** \file linalg.hpp
+ *
+ *  \brief Define some common type used in SDDK::
+ */
+
+#ifndef SIRIUS_LINALG_BASE_H
```
toxa81, line 6:
```
#ifndef __TYPE_DEFINITIONS_HPP__
...
```
and also the documentation section for the header is wrong

toxa81:
@AdhocMan can you have a look. To me it looks good

toxa81:
bors try

AdhocMan (changes requested):

toxa81:
How constexpr worked if the C++ requirement wasn't changed in CMakeLists.txt? Isn't it C++14 feature then?

src/SDDK/wf_ortho.cpp
```diff
@@ -203,34 +202,34 @@ orthogonalize(::spla::Context& spla_ctx__, memory_t mem__, linalg_t la__, int is
             /* multiplication by triangular matrix */
             for (auto& e : wfs__) {
                 /* wave functions are complex, transformation matrix is complex */
-                if (std::is_same<T, double_complex>::value) {
+                if constexpr (std::is_same<T, std::complex<precision_type>>::value) {
```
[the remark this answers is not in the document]
manhin321: I know that if constexpr is a c++17 feature, but I recall there is a compiler error if I don't use it, as the compiler require all the if branches match the function prototype to be called even the branch would not be executed in reality.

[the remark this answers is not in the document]
AdhocMan: Using C++17 features can easily cause issues with specific compilers like icc or nvcc, so this has to be removed. The issue can be resolved differently in many other ways, that don't change requirements for the entire code base.

src/SDDK/type_definition.hpp
manhin321: I haven't add the remaining copyright header as I am not sure what is the year to put in

AdhocMan (commented):
Looks good to me now. I guess we wait with merging until it's fully implemented?

bors try

toxa81:
bors try

toxa81:
> Looks good to me now. I guess we wait with merging until it's fully implemented?
> 

I suggest to merge in small increments. Otherwise we will end up in a huge PR.

PR #642, opened 2021-06-02
set of fixes for spglib interface

toxa81:


src/symmetry/rotation.hpp
```diff
@@ -197,40 +197,39 @@ inline vector3d<double> euler_angles(matrix3d<double> const& rot__)
...
-            //if (std::abs(rot__(i, j) - rit(i, j)) > 1e-8) {
-            //    std::stringstream s;
-            //    s << "rotation matrix is not unitary" << std::endl
-            //      << "initial symmetry matrix:" << std::endl
-            //      << rot__ << std::endl
-            //      << "inverse transpose matrix:" << std::endl
-            //      << rit;
-            //    RTE_THROW(s);
-            //}
+            if (std::abs(rot__(i, j) - rit(i, j)) > tolerance__) {
```
toxa81, line 200: This was a bad idea to average the input potentially non-unitary matrix. Not all unitary transformations are rotations and the problem of having a wrong symmetry here is not solved. euler_angles() should check for strict unitary matrix

toxa81:
bors try

src/symmetry/crystal_symmetry.cpp
```diff
@@ -214,26 +220,32 @@ Unit_cell_symmetry::Unit_cell_symmetry(matrix3d<double> const& lattice_vectors__
...
     if (spg_dataset_) {
+        auto lat_sym = find_lat_sym(lattice_vectors_, tolerance_);
         /* make a list of crystal symmetries */
         for (int isym = 0; isym < spg_dataset_->n_operations; isym++) {
-            auto sym_op = get_spg_sym_op(isym, spg_dataset_, lattice_vectors__, num_atoms__, positions__, tolerance__);
-            /* add symmetry operation to a list */
-            space_group_symmetry_.push_back(sym_op);
+            matrix3d<int> R(spg_dataset_->rotations[isym]);
+            for (auto& e: lat_sym) {
+                if (e == R) {
```
toxa81, line 235: That's the key point of this PR: we accept only part of SPGLIB symmetries for which the lattice is invariant under rotation with a given tolerance

toxa81:
The story is the following. Current version of SPGLIB does not fully respect it's own tolerance parameter. For "normal" lattices that's not a problem. Problems appear for slightly distorted lattices where SPGLIB tends to overestimate the number of crystal symmetries. It just doesn't notice the small distortion. The fix is to find the lattice rotation symmetries with a given tolerance ourselves and then add the SPGLIB symmetries corresponding to the found lattice symmetries. This will filter out the "bad" symmetries of SPGLIB.

simonpintarelli (approved):
lgtm

PR #644, opened 2021-06-07
print non local occupation matrix

toxa81:


toxa81:
bors try

toxa81:
* better output of nearest neighbors
* add non-local part of the occupation matrix
* print the non-local part
* test26 is the example of the non-local Hubbard input

mtaillefumier:
looks good to me so far. I saw you split the local and non local occupations.

toxa81:
Next PR will be about computing the non-local part of the occupation matrix.

mtaillefumier:
that's perfectly fine with me. I will review it.

PR #645, opened 2021-06-09
add inverse symmetry atomic list

toxa81:
store list of atoms that transform under inverse symmetry operations and also a translation vector that brings symmetry-transform atom back to the first unit cell


toxa81:
bors try

toxa81:
@mtaillefumier this is a more logical way of using symmetry-related atoms list  in the symmetrization of atomic-centered quantities

mtaillefumier:
looks good to me. I think you can merge this.

PR #646, opened 2021-06-11
Template develop

manhin321:
Template Gvec class and Wave_functions class

### DIFF
--- a/src/band/residuals.cpp
+++ b/src/band/residuals.cpp
@@ -34,7 +34,7 @@ namespace sirius {
 
 static void
 compute_residuals(sddk::memory_t mem_type__, sddk::spin_range spins__, int num_bands__, sddk::mdarray<double, 1>& eval__,
-                  sddk::Wave_functions& hpsi__, sddk::Wave_functions& opsi__, sddk::Wave_functions& res__)
+                  sddk::Wave_functions<double>& hpsi__, sddk::Wave_functions<double>& opsi__, sddk::Wave_functions<double>& res__)
 {
     for (int ispn: spins__) {
         if (is_host_memory(mem_type__)) {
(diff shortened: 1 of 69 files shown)


toxa81:
@manhin321 What is the use case for templated G-vector class? From what I see, it's only the Cartesian components and length are affected.

manhin321:
> @manhin321 What is the use case for templated G-vector class? From what I see, it's only the Cartesian components and length are affected.

It was originally intended to template everything inside the Wave_functions class to allow full fp32 calculation including the unit cell matrix storage. This change would indeed create overhead for copying when doing mix-precision calculation. 
We can still allow the templated version of Gvec class but always use Gvec<double> to see the difference later

toxa81:
Still, I want to argue that wave-functions in FP32 can co-exist with the G-vector Cartesian coordinates in FP64. You will use the G^2/2 as a kinetic energy term, and this can be contained inside Local_operator class.

manhin321:
I agree with that. I just take reference from the Julia dftk code to propose the change. Otherwise, I try to roll back for non templated Gvec class version

toxa81:
> I agree with that. I just take reference from the Julia dftk code to propose the change. Otherwise, I try to roll back for non templated Gvec class version

Yes, Gvec should be rolled back to it's original non-templated version.

src/band/residuals.cpp
```diff
@@ -34,7 +34,7 @@ namespace sirius {
 
 static void
 compute_residuals(sddk::memory_t mem_type__, sddk::spin_range spins__, int num_bands__, sddk::mdarray<double, 1>& eval__,
```
toxa81, line 36: Shouldn't it be templated over Wave_function type?

toxa81 (changes requested):
A lot of changes come from templated G-vector class. I propose to revert it back to non-templated class.

PR #648, opened 2021-06-18
Template Wave_functions class

toxa81:


toxa81:
bors try

toxa81:
This looks good to me. @AdhocMan can you also have a look when you have time?

src/SDDK/wave_functions.hpp
```diff
@@ -329,26 +335,82 @@ class Wave_functions
...
+            if (has_mt()) {
+                std::copy(src__.mt_coeffs(ispn__).prime().at(memory_t::host, 0, i0__),
+                          src__.mt_coeffs(ispn__).prime().at(memory_t::host, 0, i0__) + nmt * n__,
+                          mt_coeffs(jspn__).prime().at(memory_t::host, 0, j0__));
+            }
+            break;
+        }
+        case device_t::GPU: {
+            /* copy PW part */
+            acc::copy(pw_coeffs(jspn__).prime().at(memory_t::device, 0, j0__),
```
AdhocMan, line 366: On GPU, this is a memcpy operation. So this will not work for different types (T !=  F). A new GPU kernel for copying with type conversion might be required.

src/function3d/smooth_periodic_function.hpp
```diff
@@ -209,24 +209,25 @@ class Smooth_periodic_function
...
         auto frg_ptr = (spfft_->local_slice_size() == 0) ? nullptr : &f_rg_[0];
 
+        using precision_type = typename std::conditional<std::is_same<T, double>::value, double, float>::type;
         switch (direction__) {
             case 1: {
                 if (gvecp_->comm_ortho_fft().size() != 1) {
                     gather_f_pw_fft();
                 }
-                spfft_->backward(reinterpret_cast<double const*>(f_pw_fft_.at(sddk::memory_t::host)), SPFFT_PU_HOST);
+                spfft_->backward(reinterpret_cast<precision_type const*>(f_pw_fft_.at(sddk::memory_t::host)), SPFFT_PU_HOST);
```
AdhocMan, line 218: The transform handle type of `spfft_` will also have to be changed. `spfft::TransformFloat` instead of `spfft::Transform` for single precision. This will have to be changed in several places, like Simulation_context.

AdhocMan (changes requested):

src/function3d/smooth_periodic_function.hpp
toxa81: Good spot! I propose that we address the FFT interoperability in the next PR. This PR is a preparation for testing  inner() product in FP32. With FFTs we'll probably have to create second instance with FP32 because we will likely store two sets of wave-functions - double and float

CMakeLists.txt
```diff
@@ -82,15 +82,15 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 
 if(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU" OR ${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
-  set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g -ggdb -DDEBUG")
-  set(CMAKE_CXX_FLAGS_RELEASE "-O3 -DNDEBUG")
-  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-g -ggdb -O2")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -DSPFFT_SINGLE_PRECISION=ON")
```
AdhocMan, line 85: This should not be added to the compile flags, since this is a CMake flag, that should be used when configuring SpFFT to support single precision.

src/function3d/smooth_periodic_function.hpp
```diff
@@ -27,13 +27,31 @@
...
 
 #ifndef __SMOOTH_PERIODIC_FUNCTION_HPP__
 #define __SMOOTH_PERIODIC_FUNCTION_HPP__
 
 namespace sirius {
 
-using double_complex = std::complex<double>;
+// type traits to handle Spfft driver for different precision type
+template <typename T>
+struct SPFFT_driver {};
```
AdhocMan, line 39: This should be moved to a more suitable location, similar to the real / complex type definition header, since this might be require elsewhere as well.

AdhocMan (changes requested):
I think we should consider making support for single precision optional (at least when single precision FFTs are used). If not, we will require SpFFT with single precision enabled, which will probably require users to reinstall SpFFT as well, as this is not the default configuration. To avoid adding macros everywhere, it would be enough to only instantiate templated classes / functions for float if a macro is defined.

src/function3d/smooth_periodic_function.hpp
toxa81: there is SDDK/fft.hpp file to interface with SpFFT. It should go there

CMakeLists.txt
toxa81: Can we check inside the host code if `SPFFT_SINGLE_PRECISION` is enabled? Then we can check if we can work with FP32 FFTs or not.

AdhocMan: Yes, the CMake config file of SpFFT defines it. So after `find_package(SpFFT 1.0.2 CONFIG REQUIRED)`, `SPFFT_SINGLE_PRECISION` is available to check. It is also defined as a macro, which is included in the `spfft.hpp` header. I'd still recommend adding an option, to catch any issue if someone explicitly requires it (it also helps with package managers like Spack). The option values could be `auto`, `on` or `off`, and default to `auto`, where we enable single precision if possible.

toxa81: > I'd still recommend adding an option, 
Option on the Sirius side? For example, introduce `SIRIUS_SINGLE_PRECISION` in CMakeLists.txt?

AdhocMan: Yes. Although, I guess to keep it similar to the current options, maybe `USE_SINGLE_PRECISION`?

AdhocMan (approved):
I've added a CMake option, to control compilation with single precision, since writing CMake configs can be a bit tricky :wink:.
@toxa81 Are you happy with the new option?

Something to keep in mind for the future:
Our CI does not compile with single precision enabled. This will require an update of the Spack package, to ensure the correct SpFFT spec is selected. So we will have to check compilation manually for now.

toxa81:
Thanks Simon! `USE_FP32` variable name looks good. Manual compilation is not a problem because @manhin321 does a lot of experimental stuff. Once we see how the mixed precision works, we can introduce fp32 tests for our CI/CD.

toxa81:
bors try

PR #655, opened 2021-07-06
U_operator class

toxa81:
This is needed to separate Potential and Hamiltonian. Potential stores the Hubbard correction matrix in FP64, but the application of the Hubbard correction (U_operator) can be in FP32.


toxa81:
@manhin321 this might be a missing piece for the Hamiltonian. Before `apply_U_operator` function was a member of Hubbard class without a type template.

mtaillefumier (approved):
No objections with these changes. Seems fine to me. Not sure why the namespace hubbard is introduced though.

PR #657, opened 2021-07-09
Templating hamiltonian to work with fp32

manhin321:
Templated Hamiltonian, Local_operator, Non_local_operator class, beta_projector class
Add float type spfft driver in simulation context class


toxa81:
a naive question: can I use `static_cast` to convert complex numbers?
```
std::complex<double> a(1.0, 2.0);
std::complex<float> b = static_cast<std::complex<float>>(a);
```

manhin321:
>  I use `static_cast` to convert complex numbers?

According to cpp reference, static_cast can be used to convert between types using a combination of implicit and user-defined conversions. Which should be allowed.

At least for your example, it compiles without any error or warning

src/hamiltonian/hamiltonian_k.cpp
```diff
@@ -764,17 +763,26 @@ void Hamiltonian_k::set_fv_h_o_it(dmatrix<double_complex>& h__, dmatrix<double_c
 //==     }
 //== }
 
-template <typename T>
-void Hamiltonian_k::apply_h_s(spin_range spins__, int N__, int n__, Wave_functions<real_type<T>>& phi__,
-                               Wave_functions<real_type<T>>* hphi__, Wave_functions<real_type<T>>* sphi__)
+template <typename T> template <typename F, typename>
```
toxa81, line 766: What is T and what is F?

manhin321: > What is T and what is F?

T is the template type of Hamiltonian_k class (either float or double)
F is a further template type of the function that can real type or complex type while the precision type is same as T

toxa81: Isn't T enough? The gamma- or generic k-point case is handled by Q_operator and D_operator internally. I'll leave it for you to decide. In future, it might be a good idea to treat Gamma-point case explicitly with a template parameter (not by a real/complex data type)

manhin321: I just try to follow the code as much as original, which is templated against real and complex type as the function called the apply_non_local_d_q function which then called the inner function that depends on the type

src/density/density.cpp
```diff
@@ -47,12 +47,12 @@ Density::Density(Simulation_context& ctx__)
 
     /*  allocate charge density and magnetization on a coarse grid */
     for (int i = 0; i < ctx_.num_mag_dims() + 1; i++) {
-        rho_mag_coarse_[i] = std::unique_ptr<spf>(new spf(ctx_.spfft_coarse(), ctx_.gvec_coarse_partition()));
+        rho_mag_coarse_[i] = std::unique_ptr<spf>(new spf(ctx_.spfft_coarse<double>(), ctx_.gvec_coarse_partition()));
```
toxa81, line 50: that's an important spot: in fp32 we will sum fp32 wave-functions squared into fp32 OR fp64 density depending on the accuracy considerations

toxa81:
If I try to modify the functions to load fft data to SpFFT in this way
```
template <typename F, typename T, typename ...Args>
using enable_return = typename std::enable_if<std::is_same<typename std::result_of<F(Args...)>::type, T>::value, void>::type;


/// Load data from real-valued lambda.
template <typename F, typename T>
inline enable_return<F, T, int>
spfft_input(spfft_transform_type<T>& spfft__, F&& fr__)
{
    switch (spfft__.type()) {
        case SPFFT_TRANS_C2C: {
            auto ptr = reinterpret_cast<std::complex<T>*>(spfft__.space_domain_data(SPFFT_PU_HOST));
            #pragma omp parallel for schedule(static)
            for (int i = 0; i < spfft__.local_slice_size(); i++) {
                ptr[i] = std::complex<T>(fr__(i), 0.0);
            }
            break;
        }
        case SPFFT_TRANS_R2C: {
            auto ptr = reinterpret_cast<T*>(spfft__.space_domain_data(SPFFT_PU_HOST));
            #pragma omp parallel for schedule(static)
            for (int i = 0; i < spfft__.local_slice_size(); i++) {
                ptr[i] = fr__(i);
            }
            break;
        }
        default: {
            throw std::runtime_error("wrong spfft type");
        }
    }
}

/// Load data from complex-valued lambda.
template <typename F, typename T>
inline enable_return<F, std::complex<T>, int>
spfft_input(spfft_transform_type<T>& spfft__, F&& fr__)
{
    switch (spfft__.type()) {
        case SPFFT_TRANS_C2C: {
            auto ptr = reinterpret_cast<std::complex<T>*>(spfft__.space_domain_data(SPFFT_PU_HOST));
            #pragma omp parallel for schedule(static)
            for (int i = 0; i < spfft__.local_slice_size(); i++) {
                ptr[i] = fr__(i);
            }
            break;
        }
        case SPFFT_TRANS_R2C: {
        }
        default: {
            throw std::runtime_error("wrong spfft type");
        }
    }
}

/// Input CPU data to CPU buffer of SpFFT.
template <typename T>
inline void spfft_input(spfft_transform_type<T>& spfft__, T const* data__)
{
    spfft_input(spfft__, [&](int ir){return data__[ir];});
}
```
I get a brutal error
```
/Users/antonk/src/github/electronic-structure/SIRIUS/src/SDDK/fft.hpp:137:5: error: no matching function for call to 'spfft_input'
    spfft_input(spfft__, [&](int ir){return data__[ir];});
    ^~~~~~~~~~~
/Users/antonk/src/github/electronic-structure/SIRIUS/src/function3d/smooth_periodic_function.hpp:227:17: note: in instantiation of function template specialization 'spfft_input<d
ouble>' requested here
                spfft_input(*spfft_, frg_ptr);
                ^
/Users/antonk/src/github/electronic-structure/SIRIUS/src/density/density.hpp:304:27: note: in instantiation of member function 'sirius::Smooth_periodic_function<double>::fft_tran
sform' requested here
        rho_pseudo_core_->fft_transform(1);
                          ^
/Users/antonk/src/github/electronic-structure/SIRIUS/src/SDDK/fft.hpp:86:1: note: candidate template ignored: couldn't infer template argument 'T'
spfft_input(spfft_transform_type<T>& spfft__, F&& fr__)
^
/Users/antonk/src/github/electronic-structure/SIRIUS/src/SDDK/fft.hpp:114:1: note: candidate template ignored: couldn't infer template argument 'T'
spfft_input(spfft_transform_type<T>& spfft__, F&& fr__)
^
/Users/antonk/src/github/electronic-structure/SIRIUS/src/SDDK/fft.hpp:135:13: note: candidate template ignored: could not match 'const T *' against '(lambda at /Users/antonk/src/
github/electronic-structure/SIRIUS/src/SDDK/fft.hpp:137:26)'
inline void spfft_input(spfft_transform_type<T>& spfft__, T const* data__)
            ^
1 error generated.
make[2]: *** [src/CMakeFiles/sirius.dir/hamiltonian/hamiltonian.cpp.o] Error 1
```

@AdhocMan can you give a word of a wisdom? How

src/density/density.cpp
manhin321: Would it sound more complete to introduce the required change in this PR? Looks like it would either require to template the class and function, or include many more variables to the class

toxa81: We will need to have summation of charge density in FP64 from the FP32 wave-functions to understand the accuracy. So let's try to keep Density and Potential in FP64.

toxa81:
This is some template sorcery @AdhocMan . I use one explicit template parameter for data type and then spfft_input(lambda) wrorks

src/k_point/k_point.hpp
```diff
@@ -60,6 +60,10 @@ class K_point
 
     std::unique_ptr<spfft::Transform> spfft_transform_;
 
+    /// A float type fft transform to supprt fp32 calculation
+#ifdef USE_FP32
+    std::unique_ptr<spfft::TransformFloat> spfft_transformFloat_;
```
toxa81, line 65: Shouldn't it go into K-point PR?

src/gpu/create_beta_gk.cu
```diff
@@ -36,7 +36,7 @@ __global__ void create_beta_gk_gpu_kernel
     int const* beta_desc__,
     gpu_complex_type<T> const* beta_gk_t,
     T const* gkvec,
-    T const* atom_pos,
+    double const* atom_pos,
```
toxa81, line 39: @manhin321 can you also check the G+k vector coordinates are also always in `double`?

src/k_point/k_point.hpp
manhin321: It was indeed a previous commit that intent not to template the k_point class. I remove it now, but keep in mind that the templated hamiltonian class for this is not ready for a float type instantiation as the spfft driver is incorrect there

toxa81: Got it! We will first merge the K_point PR and then resolve the conflicts here.

PR #658, opened 2021-07-13
add a template parameter

toxa81:
The purpose of this PR is to introduce a dependency of K_point over a data-type to store FP64 or FP32 wave-functions.


toxa81:
The K_point_set class will now store fp64 and fp32 K_points with wave-functions. Later it will be decides if only one set is enough.

PR #661, opened 2021-07-14
Methfessel Paxton smearing / Newton minimization

simonpintarelli:
add Methfessel-Paxton smearing and Newton minimization protocol as proposed in https://gitlab.com/QEF/q-e/-/merge_requests/1294


toxa81 (approved):
all looks good

Issue #662, opened 2021-07-16
API: advertised parameters not available anymore


dev-zero:
Essentially the SIRIUS side of https://github.com/cp2k/cp2k/issues/1590

```
=== Warning at line 104 of file /data/tiziano/cp2k/tools/toolchain/build/SIRIUS-7.2.5/src/context/simulation_parameters.cpp ===
The following configuration parameters were not recognized and ignored: potential_tol
```

It seems that the options were changed from 7.0 to 7.2 and neither the `options.json` nor the verification tests or examples were updated. I started looking into how to fix this quickly, but I wonder why there are two divergent JSON schemas `options.json` and `input_schema.json`.

toxa81:
Hi Tiziano! `input_schema.json` is a clean re-implementation of `options.json`. In future we plan to remove `options.json`, but to do so a few API functions that work with CP2K have to be updates as well.  Inside SIRIUS only  `input_schema.json` is used.

dev-zero:
@toxa81 ok, thanks. Any ETA on that? Some options (like `potential_tol`) can currently not be set from CP2K (with release versions of CP2K and SIRIUS).

toxa81:
We need to ask @mtaillefumier for a help.

mtaillefumier:
I am off the all week but I will look into it. 

I can say that we changed some option names (the option schema Anton introduced recently). I am working on using the option schema for the cp2k input parameters. I did not finish this work though (I was hopping to finish it last week). It will need a PR and new version of SIRIUS and a PR for cp2k.

toxa81:
@mtaillefumier @dev-zero @oschuett  
are we OK to close the issue? I think this is polished now

dev-zero:
I think so, yes. Works for me with the current CP2K master and SIRIUS v7.2.8

PR #663, opened 2021-07-16
generate_api: make it more user friendly

dev-zero:


dev-zero:
while looking into a fix for #662 I updated the generate_api command a bit since it simply failed on me

toxa81:
Thanks @dev-zero ! All looks good to me.

PR #668, opened 2021-07-26
[Fortran interface] also use input_schema.json for retrieving input parameters

mtaillefumier:
the fortran interface for requiring SIRIUS input parameters now uses Anton json dictionary. It means that the original option dictionary is obsolete and removed from the source code. It implies that  options that can take empty strings as default (with finite number of option values) now take "auto" as default. The default behavior was empty string = auto is preserved.

We might need to update the minor version of SIRIUS because it is needed for cp2k.

there is no api break so far.


simonpintarelli (approved):
lgtm

src/context/simulation_parameters.cpp
```diff
@@ -186,46 +203,48 @@ void Simulation_parameters::processing_unit(std::string name__)
...
-                    for (size_t st = 1; st < v.size(); st++)
-                        std::cout << " " << v[st];
-                }
-                std::cout << "default value : " << dict[el.key()]["default_values"].get<std::string>() << std::endl;
-            }
-        }
-    }
-    Communicator::world().barrier();
-}
+// void Simulation_parameters::print_options() const
```
AdhocMan, line 213: Maybe this part could be removed instead?

AdhocMan (approved):
LGTM overall.

src/context/simulation_parameters.cpp
mtaillefumier: I may have to restore the next function. I will look into it. Thanks :)

PR #669, opened 2021-07-29
[API] Use template programming for extracting and setting options

mtaillefumier:
- Fix a bug in the interface. CP2K was not happy at all by the things I added. Needed fixing
- Use template function to remove code duplication. Add to specialize for string.


simonpintarelli (approved):
lgtm

PR #671, opened 2021-08-11
relax a bit threshold on the symmetry check

toxa81:
In the LAPW high-cutoff mode, the symmetry check was too tough as reported by @dev-zero


toxa81:
bors try

PR #674, opened 2021-08-20
fix the way nlcg is set up

toxa81:
Hi @simonpintarelli 
For some reason, the NLGC options were not updated to be fetched from the new Config() parameters.


simonpintarelli:
@toxa81 Thanks, yes indeed. My PR will come soon.

PR #675, opened 2021-08-23
emplace_back XC_functional(), do not use pointers

toxa81:
Also, remove insert_xc_functional() from Potential class, as it is not used.


toxa81:
bors try

toxa81:
bors try

simonpintarelli (approved):
lgtm

PR #678, opened 2021-09-05
Template GPU Eigensolver

manhin321:
Scalapack and cusolver is templated to enable fp32


src/linalg/eigenproblem.hpp
```diff
@@ -827,9 +827,19 @@ class Eigensolver_scalapack : public Eigensolver
     }
 
     /// Solve a standard eigen-value problem for all eigen-pairs.
-    int solve(ftn_int matrix_size__, dmatrix<double_complex>& A__, double* eval__, dmatrix<double_complex>& Z__)
+    template <typename T, typename = std::enable_if_t<!std::is_scalar<T>::value>>
+    int solve_(ftn_int matrix_size__, dmatrix<T>& A__, real_type<T>* eval__, dmatrix<T>& Z__)
     {
-        PROFILE("Eigensolver_scalapack|pzheevd");
+        if (std::is_same<T, std::complex<double>>::value) {
+            PROFILE("Eigensolver_scalapack|pzheevd");
```
toxa81, line 834: This won't work, because PROFILE macro lives inside a scope and will be destroyed afterwards. Better to do ```PROFILE("Eigensolver_scalapack|pxheevd");``` at the beginning of the function

manhin321: I pull it out to the wrapper function now, see if this is satisfied, although it may introduce a little bit WET to the code

PR #680, opened 2021-09-14
Fix a bug in the api. Options taking string parameters are handled in…

mtaillefumier:
- add the electronic structure method parameter (optional) to sirius_set_parameters. It should not break the api in any way
- Fix a mistake in the function setting runtime options that take string as parameter

Can we bump a new minor revision when this patch is merged ?


toxa81:
@mtaillefumier we can merge to develop and tag 7.2.7. According to our internal policy, we don't do direct PRs to 'master'.

toxa81:
@haampie is ROCm package outdated?

PR #681, opened 2021-09-14
try to remove linearly dependent states

toxa81:
Try to get rid of the linearly-dependet basis functions during the orthogonalization phase.


toxa81:
I introduced a debug variable `project_out_here`. If` true`, the projection and filtering of new basis functions (preconditioned residuals) happens in davidson(). If `false` - in wf_ortho()`. They should be identical! However, they are clearly not. In first case the davidson fails to converge, in the second (which is a default for the main branch) - it converges. To test, run `test_davidson --num_bands=8`

PR #709, opened 2021-12-29
upf2_to_json: cutoff_radius_index is optional for non-local part

dev-zero:
Not entirely sure about the auto-detection of the cutoff-index.
Meaning, this is a format conversion tool, and while this might be useful,
a user might not expect it?


toxa81:
This looks fine to me. Unfortunately, there is no schema for the UPF files. Hence, the conversion needs to handle all the possible flavors of the upf file.

PR #710, opened 2022-01-03
single file for Fortran API

toxa81:
Modify python script to write a single `sirius.f90` file with Fortran API.


mtaillefumier (approved):

mtaillefumier:
Simplifications are always welcome. Do you want me to test with cp2k ?

toxa81:
> Simplifications are always welcome. Do you want me to test with cp2k ?

It should not brake anything, but a simple test is welcomed.

mtaillefumier:
I will do it then.

mtaillefumier:
it is good to go. Compilation and runtests passed without any problem.

PR #713, opened 2022-01-04
yet another change in ELPA api

toxa81:
ELPA has changes the API and we need to adjust. The main concern is CP2K - do they support elpa-2021-11-001 already?


mtaillefumier:
Yes cp2k supports 2021.11.001. no risk of breaking anything on cp2k side.

toxa81:
@haampie why CI doesn't pick latest ELPA to install with spack?

toxa81:
bors try

PR #723, opened 2022-02-07
Develop

mtaillefumier:
some cp2k issue popping up that imply a fix in sirius.


toxa81:
bors try

toxa81:
I think input parser should be re-generated from json schema

PR #724, opened 2022-02-09
[CI] add view /root/local for the CPU build

toxa81:
Having this in Dockerfile is awkward:
```
RUN echo "export SIRIUS_DIR=$(spack location -i sirius@develop%gcc)" >> /etc/profile.d/sirius.sh
```
Try to install sirius+dependencies into /root/local

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     container: stabbles/sirius-cpu
     env:
-      SPEC: sirius@develop %gcc +tests +apps +vdwxc +fortran build_type=RelWithDebInfo ^netlib-lapack ~external-blas ^mpich
+      SPEC: sirius@develop %gcc +tests +apps +vdwxc +fortran build_type=RelWithDebInfo ^intel-mkl@2017: threads=openmp ^mpich
     steps:
       - uses: actions/checkout@v2
       - name: Setup Spack
@@ -84,7 +84,7 @@ jobs:
     runs-on: ubuntu-latest
     container: stabbles/sirius-cpu
     env:
-      SPEC: sirius@develop %gcc +vdwxc +tests +apps +elpa build_type=RelWithDebInfo ^netlib-lapack ~external-blas ^mpich
+      SPEC: sirius@develop %gcc +vdwxc +tests +apps +elpa build_type=RelWithDebInfo ^openblas ^mpich
     steps:
       - uses: actions/checkout@v2
       - name: Setup Spack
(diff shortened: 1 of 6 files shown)


ci/spack/cpu.yaml
```diff
@@ -3,7 +3,7 @@ spack:
     - sirius@develop %gcc +vdwxc +tests +apps build_type=RelWithDebInfo +elpa ^openblas %gcc ^mpich
     - sirius@develop %gcc +vdwxc +tests +apps build_type=RelWithDebInfo +elpa ^openblas %gcc ^openmpi
     - sirius@develop %clang +tests +apps build_type=RelWithDebInfo ^openblas ^mpich ~fortran
-  view: false
```
haampie, line 6: I don't think this will work, cause there's multiple flavors of sirius being built (%gcc, %clang, ^openmpi, ^mpich), so they will clash with filenames. You can do something like

```
view:
  gcc-with-mpich:
    root:/root/local/gcc-with-mpich
    select: ['%gcc ^mpich']
  gcc-with-openmpi:
    root:/root/local/gcc-with-openmpi
    select: ['%gcc ^openmpi']
  clang-with-mpich:
    root:/root/local/clang-with-mpich
    select: ['%clang ^openmpi']
```

to avoid the clashes, but that seems a bit verbose

toxa81:
@haampie I will use this PR to also fix the openblas->netlib blas

toxa81:
This looks better. I disabled rocm builds until we get a stable ROCm PE back. For the cpu runs MKL is now used and voila! test15 passes

.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     container: stabbles/sirius-cpu
     env:
-      SPEC: sirius@develop %gcc +tests +apps +vdwxc +fortran build_type=RelWithDebInfo ^openblas ^mpich
+      SPEC: "sirius@develop %gcc +tests +apps +scalapack +fortran build_type=RelWithDebInfo ^intel-mkl@2017: threads=openmp ^mpich"
```
haampie, line 14: I think `threads=openmp` is not doing anything

toxa81: I think it was borrowed from some example in the past. Let's try without threads=openmp

toxa81: works

haampie: FWIW, rocm should also be back

haampie: oh it's already merged :laughing:

PR #725, opened 2022-02-11
Remove incomplete G-vector shells

toxa81:
Selecting G-vectors by cutoff radius is not stable. In the rare cases we might miss the G-vectors which belong to the shell with a cutoff radius, but due to numerical noise have a slightly larger radius. In this case we are generating incomplete G-vector shell. The fix searches for the incomplete shells and removes such G-vectors.


toxa81:
bors try

toxa81:
bors try

toxa81:
bors try

mtaillefumier (approved):
I do not see anything wrong here. The G-vectors of a given shells are generated externally. We use symmetry to check that the present G-vectors have all their image by symmetry or delete them if not. 
The code is relevant for the outer shell but it is always good to check the inner shells as well. I would probably reject a incomplete shell all together or make it complete.

toxa81:
Current code rejects the incomplete shell. So all the G-vectors will have length <= G_cutoff

PR #726, opened 2022-02-17
Uplus v2 from @mtaillefumier

toxa81:
This PR includes all necessary bit of code to run simulations with the LDA+U+V correction and more. It includes so far

- Multi-orbital case for both U and V corrections
- Possibility to orthogonalize the hubbard subspace compared to all atomic wave functions
- Forces (for U+V) and stresses for the co-linear case
- 3 more tests for testing the code

Remains to be done

- [ ] Move symmetrization of the occupations matrices to the right place (the function would need too many parameters, so need a proper fix for this)
- [x] Add a function for adding atom pairs for the V term in the api
- [x] Extra work on QE side
- [ ] And of course more testing
- [x] remove duplicated code for finding the orbital index

The multi-orbital case is treated such that the code keep a list of "atomic" levels to be considered in the U (and / or V) term. So instead of having occupations numbers per atoms we have occupations numbers per atomic levels. This would be equivalent to having a full matrix that we decompose per block, each block representing a term in the correction.


toxa81:
@mtaillefumier I fixed all the merge issues. Let's see how ci will run it.

toxa81:
> The V correction is still finicky but the U part should give you the answer than QE provided we use the same functional etc... I was planning to work it but of course things always get delayed.

I offer a trade :) I'll massage this a little bit, and you have a look at the forces/stress in case of SO. There is something fishy there. There is no explicit dependency on SO switch, yet it gives a different answer comparing to QE.

mtaillefumier:
Feel free to do it. My guess is that the phase factors are wrong (some conjugate wrongly placed). QE code is difficult or impossible to decipher on the top of it. 
So : Let discuss that tomorrow quickly.

toxa81:
bors try

toxa81:
bors try

toxa81:
bors try

toxa81:
bors try

toxa81:
bors try

src/density/occupation_matrix.hpp
```diff
@@ -45,18 +47,38 @@ class Occupation_matrix : public Hubbard_matrix {
...
+                                      static_cast<int>(this->local(at_lvl).size()));
             }
         }
+
+        // we must reduce occ_mtrx_T_ not nonlocal (it non zero only after symmetrization)
+        for (auto& T : this->occ_mtrx_T_) {
+            ctx_.comm().allreduce(T.second.at(memory_t::host), static_cast<int>(T.second.size()));
+        }
+
+        // std::cout  << "Before : " << this->nonlocal(0)(0, 0, 0) << std::endl;
```
mtaillefumier, line 68: these are debugging statements I forgot to remove. We can most probably just delete them.

src/hamiltonian/non_local_operator.cpp
```diff
@@ -414,67 +418,79 @@ void
...
-                        // TODO: replace this with matrix matrix multiplication
-
+                      // TODO: replace this with matrix matrix multiplication
                         for (int nbd = 0; nbd < n__; nbd++) {
                             for (int m1 = 0; m1 < lmax_at; m1++) {
                                 for (int m2 = 0; m2 < lmax_at; m2++) {
-                                    Up(um__.nhwf() * s1 + um__.offset(ia) + m1, nbd) +=
-                                        um__(um__.offset(ia) + m2, um__.offset(ia) + m1, ind) *
-                                        dm(um__.nhwf() * s2 + um__.offset(ia) + m2, nbd);
+                                    const int ind = (s1 == s2) * s1 + (1 + 2 * s2 + s1) * (s1 != s2);
```
mtaillefumier, line 472: It is probably worth to have a small macro or inlned function the sirius namespace for computing these indices. It is not blocking though and can wait.

mtaillefumier (approved):
We should probably merge this in develop as U correction works. 

Possible places or errors
- mixing of the non local occupations
- k-point reduction of the non local parts (I checked the code and it sounds just fine)
- sign in energy.cpp (line 181 should be minus as it is already)

toxa81:
bors try

toxa81:
> ## try
> 
> Already running a review

@haampie Do you know how to force bors to kill the hanging job and to re-run the pipeline?

haampie:
bors try

toxa81:
bors try

toxa81:
bors try

PR #729, opened 2022-04-06
Fix the symmetrization of non-local occupancy matrix

toxa81:
This should fix the non-local non-orthoatomic LDA+U+V case.


src/SDDK/geometry3d.hpp
```diff
@@ -506,13 +506,17 @@ inline std::ostream& operator<<(std::ostream& out, matrix3d<T> const& v)
     return out;
 }
 
-inline std::pair<vector3d<double>, vector3d<int>> reduce_coordinates(vector3d<double> coord)
+/// Reduce the coordinates to the first unit cell.
+/** Split the input vector in lattice coordinates to the sum r0 + T, where T is the lattice ranslation
```
mtaillefumier, line 510: typo

mtaillefumier (approved):
the std::real was the issue ?

PR #732, opened 2022-05-04
[enhancement] Use C++ streams for output

toxa81:
Background: the usage of printf() was inspired by the Google C++ coding standards (https://google.github.io/styleguide/cppguide.html) which _now_ don't cotain a recommendation to use printf(). The usage of printf() should be gradually removed.
In this PR:
 * introduce and start using output stream in the simulation context
TODO:
 - [x] write checksums and hashes to std::cout, they will not be captured in the output stream
 - [ ] better place (source file) for output-related functionality


mtaillefumier (approved):
merge when you see fit :)

simonpintarelli:
It doesn't make a difference, since it is not used in the code. But `pstdout << "test"` will return an object of `std::stringstream` and the same for `rte::ostream`. I think not inheriting from `std::stringstream` would be nicer and instead make these objects private members of the class and add `pstdou& operator<<(const std::string&)` as member function. 
Is `pstdout::get` needed? E.g. can't it be included in `pstdout::flush`?

toxa81:
I can make the changes according to your commens. The only caviat is that `pstdou& operator<<(const std::string&)` is not enough, thare are format modifiers and things like std::endl. They also need to be handeled in the operator overload.

simonpintarelli:
> I can make the changes according to your commens. The only caviat is that `pstdou& operator<<(const std::string&)` is not enough, thare are format modifiers and things like std::endl. They also need to be handeled in the operator overload.

Sorry, you're right. Then perhaps it is best to leave it as it is, adding all those functionalities might be cumbersome.

toxa81:
@simonpintarelli We can discuss what you proposed in this PR: https://github.com/electronic-structure/SIRIUS/pull/785

PR #735, opened 2022-05-11
Add hubbard ortho case for forces

mtaillefumier:
- Hubbard forces for U term work as expected. Sirius gives the same answer than QE. 
- Need to check for the V term but computation of the occupancies derivatives is identical. So the only possible place where things could go wrong is in forces.cpp. 

we can reduce computing cost even further at the prize of memory


toxa81:
@mtaillefumier can you please do `git push -u origin your-branch-name` to push your branch to this repository and then open a PR.

mtaillefumier:
I can apply the same commit on any branch you wish.

PR #745, opened 2022-07-04
fix compilation with gcc12

simonpintarelli:
Default template argument was defined twice in definition and declaration. Remove it from the definition.


toxa81:
Hi @simonpintarelli ! If you can review this PR https://github.com/electronic-structure/SIRIUS/pull/733 instead of Harmen (who is on vacation), the same fix is applied there.

simonpintarelli:
Sure. I'll close this PR then.

PR #747, opened 2022-07-04
New wave-functions implementation

toxa81:
Refactor wave-functions class and functionality around Hamiltonian application. The summary of the changes is the following:

1. Old Wave_function class is replaced by Wave_function, Wave_function_fft and Wave_function_mt classes. The concept of "extra" storage is removed. The transformation between data layouts is done by COSTA. There is still one caveat: the performance of COSTA library is low (https://github.com/eth-cscs/COSTA/issues/10) and that's why the custom code to shuffle wave-functions between slab and FFT representation is kept. The plane-wave and muffin-tin coefficients are now stored in one consecutive array. This allows to simplify the inner() and transform() functions and at the same time avoid two calls to SPLA and thus reduce communication.
2. Two template parameters T and F are now used in all functions that work with Hamiltonian or wave-functions. T is the precision type of wave-functions (double or float) and F is the type of the Hamiltonian matrix: float or double for Gamma-point case, complex\<float\> or complex\<double\> for general k-point case.
3. Strong types for indices are introduced. Now instead of a foo(int, int) we have something more meaningful like foo(wf::spin_index, wf::band_index). This should be adopted in other parts of the code
4. inner() function of beta-projectors is replaced by the common wf::inner() function
5. mstrix_storage class is removed as deprecated
6. memory_usage parameter is removed from json schema as deprecated
7. some CUDA kernels are templated; common cuda functions are introduced in src/gpu/acc_common.hpp to work with real and complex number

TODO:
 - [x] @toxa81 restore all tests and cleanup unused
 - [x] @toxa81 write a helper function for the tests of hloc and davidson to create fplapw or pp-pw simulation context
 - [x] @toxa81 restore mixed FP32/FP64 wf::inner() and wf::transfrom() 
 - [x] @toxa81 add magma ci/cd
 - [x] @simonpintarelli restore the interface to NLCG
 - [x] check test02, it doesn't work ATM
 - [x] restore Fortran API

@mtaillefumier @simonpintarelli Guys! My apologies for this monster PR. Just have a bird-eye overview. That would be enough.


toxa81:
bors try

toxa81:
This CI/CD can be tricky here. New code used simple features of c++17. We first need to upgrade the base image to cuda11 to switch to c++17.

toxa81:
bors try

mtaillefumier:
no worries. I will have a look at it

mtaillefumier (approved):
that's a big PR. I went through it and the minor issues I have are mostly related to commented dead code. It is by far not blocking.
We will need active testing considering the size of the PR.

apps/tests/test_examples.cpp
```diff
@@ -106,39 +106,39 @@ double Gmax = 10;
...
-inner(spla_ctx, spin_range(ispn), wf, 0, N, wf, 0, N, o, 0, 0);
-/* get the diagonal of the matrix */
-auto d = o.get_diag(N);
-/* check diagonal */
-for (int i = 0; i < N; i++) {
-    if (std::abs(d[i] - 1.0) > 1e-10) {
-        throw std::runtime_error("wrong overlap");
-    }
-}
+///* create G-vector partition; second communicator
```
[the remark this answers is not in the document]
toxa81: it's actually in my TODO to restore it back

src/gpu/residuals_aux.cu
```diff
@@ -418,18 +529,83 @@ extern "C" void make_real_g0_gpu_float(acc_complex_float_t* res__,
...
+            y__[k2] = mul_accNumbers(alpha, x__[k1]);
+        } else if (is_zero(alpha)) {
+            y__[k2] = mul_accNumbers(beta, y__[k2]);
+        } else {
+            y__[k2] = add_accNumbers(mul_accNumbers(alpha, x__[k1]), mul_accNumbers(beta, y__[k2]));
+        }
+    }
+}
 
+extern "C" {
```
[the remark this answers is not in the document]
toxa81: this should be fine – it is just a pointer to some type and no overloading is happening here

simonpintarelli (approved):
lgtm

src/SDDK/wave_functions.hpp
```diff
@@ -27,482 +27,1970 @@
...
+ * \f]
+ * where i0 and j0 and the dimensions of the resulting inner product matrix are determined by the band ranges for
+ * bra- and ket- states.
+ *
+ * The location of the wave-functions data is determined by the mem parameter. The result is always returned in the
+ * CPU memory. If resulting matrix is allocated on the GPU memory, the result is copied to GPU as well.
+ */
+template <typename T, typename F, typename W>
+inline std::enable_if_t<std::is_same<T, real_type<F>>::value, void>
+inner(::spla::Context& spla_ctx__, sddk::memory_t mem__, spin_range spins__, W const& wf_i__,
```
simonpintarelli, line 1539: Would be nice to generalize this to arbitrary arguments for beta-projectors and wave-functions. But we can also try to this after the PR is merged.

toxa81: Do you mean for any wave-function like objects (both <bra| and |ket>)?

simonpintarelli: Yes, I think that would be the best, to have a template argument for both bra and ket.

PR #748, opened 2022-07-04
Cleanup unused variables / Warnings

simonpintarelli:
Fix warnings, unused variables. Remove unnecessary temporary variables followed by std::move.
Replace `const Communicator&` -> `Communicator` in `Gvec` class.


toxa81:
bors try

PR #767, opened 2022-09-13
new ci workflow

toxa81:
Switching to base containers that are uploaded to electronicstructure account on Dockerhub and generated by us. The actual Dockerfiles will follow.


simonpintarelli (commented):
lgtm

PR #769, opened 2022-09-23
Add rocsolver for symmetric/hermitian eigenvalue problem

simonpintarelli:


simonpintarelli:
the rocsolver wrappers have been merged with the nlcglib pr.

PR #772, opened 2022-10-05
Change API to be able to pass array(:,:,:) kind of objects

toxa81:
There is also a fix for calculation of IBS forces in LAPW. The radial wave functions should not be recomputed.


toxa81:
Hi @mtaillefumier ! This might break the cp2k interface in favor of a more Fortran-ish style. I'm testing it now with Exciting and QE.

PR #775, opened 2022-10-20
[fix] Adjust to the latest ELPA API

toxa81:


toxa81:
I will merge this before the ELPA spack recipie is updated. We need to tag a new release asap.

PR #777, opened 2022-11-04
WIP: update api for python module, drop python2

simonpintarelli:
- [x] drop python2
- [x] update to new SIRIUS API for most important parts
- [ ] cleanup / split `py_sirius.cpp` into multiple cpp files 
- [x] @toxa81  ci/cd pipeline for python module


spack/packages/sirius/package.py
```diff
@@ -95,7 +95,7 @@ class Sirius(CMakePackage, CudaPackage):
     depends_on('pkgconfig', type='build')
 
     # Python module
-    depends_on('python', when='+python', type=('build', 'run'))
+    depends_on('python', when='+python@3.6:', type=('build', 'run'))
```
toxa81, line 98: IMO we should drop this file as well and submit changes directly to the spack package of sirius

toxa81:
The vector3d<double> intrface is not working properly.   This works but is awkward: 
https://github.com/electronic-structure/SIRIUS/blob/update/python-module/examples/python/Si_dft.py#L29

This should be the intrface (https://github.com/electronic-structure/SIRIUS/blob/update/python-module/examples/python/Si_dft.py#L30), but it gives the error: 

```
TypeError: add_atom(): incompatible function arguments. The following argument types are supported:
    1. (self: sirius.py_sirius.Unit_cell, arg0: str, arg1: geometry3d::vector3d<double>) -> None
```

simonpintarelli:
> The vector3d intrface is not working properly. This works but is awkward: https://github.com/electronic-structure/SIRIUS/blob/update/python-module/examples/python/Si_dft.py#L29
> 
> This should be the intrface (https://github.com/electronic-structure/SIRIUS/blob/update/python-module/examples/python/Si_dft.py#L30), but it gives the error:
> 
> ```
> TypeError: add_atom(): incompatible function arguments. The following argument types are supported:
>     1. (self: sirius.py_sirius.Unit_cell, arg0: str, arg1: geometry3d::vector3d<double>) -> None
> ```

I'll have a look.

simonpintarelli:
@toxa81 Now both variants `Unit_cell.add_atom(label, [list_like])` and `Unit_cell.ad_atom(label, sirius.vector3d)`. The former also does check the length, it will accept only python list / numpy arrays with 3 entries (pybind11 does that).

toxa81:
This looks good to me now. Unless @simonpintarelli you _really_ want to work on `cleanup / split py_sirius.cpp into multiple cpp files`, it is good to go for meging. The Python intrface is restored to a minimum that can run DFT ground state.

simonpintarelli:
@toxa81 Let's merge it. I'll work on it later.

PR #778, opened 2022-11-07
[feature] Implementation of the interface to geomtry optimization.

toxa81:
Hi @moritzgubler!
I interfaced your vc-sqnm library. All works fine. Now I need to figure out a few things:
 - what are the cryptic numerical parameters to the constructor of periodic_optimiser `(na, lat_a, lat_b, lat_c, 2.0, 10, 2.0, 1.e-2, 1.e-4);`
 - what is the stop criteria of the library when the lattice is converged? how do I set it? 
 - for equlibrium geomery (when forces are zero) the atomic optimiser will crash

TODO:
 - [x] add input parameters to the json schema
 - [x] add tolerance for forces and stress


moritzgubler:
Hi @toxa81 

- The src/cpp/periodic_optimizer.hpp class can do two tasks. Fixed cell geometry optimization and variable cell shape geometry optimization depending on which constructor is used to create the object. For variable cell shape geometry optimization (and fixed cell) there are two constructors, one version that allows you to set all parameters and one where you only have to provide the absolute minimum of parameters and default values are used for the rest of them. The exact meaning of all the parameters in documented in the src/cpp/periodic_optimizer.hpp file. In the case of the constructor you used:
        /**
     * @brief Construct a new periodic optimizer::periodic optimizer object for variable cell shape optimization with custom parameters.
     * 
     * @param nat number of atoms
     * @param lat_a first lattice vector
     * @param lat_b second lattice vector
     * @param lat_c third lattice vector
    * @param initial_step_size initial step size. default is 1.0. For systems with hard bonds (e.g. C-C) use a value between and 1.0 and
    * 2.5. If a system only contains weaker bonds a value up to 5.0 may speed up the convergence.
    * @param nhist_max Maximal number of steps that will be stored in the history list. Use a value between 3 and 20. Must be <= than 3*nat.
    * @param lattice_weight weight / size of the supercell that is used to transform lattice derivatives. Use a value between 1 and 2. Default is 2.
    * @param alpha0 Lower limit on the step size. 1.e-2 is the default.
    * @param eps_subsp Lower limit on linear dependencies of basis vectors in history list. Default 1.e-4.

You can use the default value for all of the parameters except for the initial step size (You can also use the default for the initial step size but it is not ideal). As explained above it is a bit system specific and would ideally be set by the user. In section 6.4 of the paper (https://arxiv.org/pdf/2206.07339.pdf) I present a simple recipe that can be used to estimate it. I did not yet have time to integrate it into the library.

- There is no stopping criterion implemented in my repository. You should loop over the periodic_optimizer.step() function until the norm of the forces and stresses are sufficiently small. That way, the vc-sqnm library never needs to be able to calculate energies and forces on its own.
- This is a bug, thanks for finding it. At some point I need to divide a quantity by the norm of the forces and I do not check if they are zero. Where the stresses also exactly zero? Or did you use the fixed cell geometry constructor? When the variable cell shape constructor was used and only the forces were exactly zero this should not have happened. I can add a mechanism that returns the geometry without any change if the forces are exactly zero.

toxa81:
@moritzgubler thanks for poiniting out to the documentation! very useful. I will need to introduce this parameterers in our input schema. (2) is clear, I will introduce the convergence parameter myself; for (3) - yes, it was fixed cell atomic relaxation.

moritzgubler:
@toxa81 In my implementation I did not check the stress tensor but the lattice derivatives. The two quantities are closely related, it probably does not matter that much which one you choose. You can calculate the lattice derivatives using:
dE / A = -det(A) * stress A^{-1}^T
where A is the 3*3 matrix containing the three lattice vectors. The lattice vectors are stored column wise in A. To check the convergence I looked at the maximal value in magnitude of the forces and lattice derivatives. In hartree / bohr 10 ^-3 is a rather loose convergence criterion and 10^-4 is probably tight enough for most applications.

In the file periodic_optimizer.hpp on line 250 there is a function that converts the stress tensor to lattice derivatives.

toxa81:
@moritzgubler  a few updates for this PR:
1) now you can set the output stream to be a file:
```
str_val = "file:output.txt"
call sirius_option_set(handler, "control", "output", SIRIUS_STRING_TYPE, C_LOC(str_val), len(trim(adjustl(str_val))))
```
2) sirius_find_ground_state() API function now contains new optional parameter `rho_min`. If the calculation is converged, you can check for rho_min. If it is negative, that is bad and total energy/forces/stress can't be trusted,

moritzgubler:
@toxa81 That's great, I think both features will be useful. 
Do you already use the new mechanism from the vc-SQNM library that estimates the initial step size? This can be done by setting the initial step size parameter to a small negative number. In hartree and bohr units, -0.1 would be a good choice.

toxa81:
> In hartree and bohr units, -0.1 would be a good choice.
I will use it as a default parameter in the json input schema.

mtaillefumier (approved):

PR #779, opened 2022-11-09
[fix] performance in ortgonalize()

toxa81:
Surprisingly, the cuda trmm function can take a cpu pointer. This was impacting the performance of the  wf::ortgonalize() function. Fixed now.


mtaillefumier (approved):
good to go

PR #783, opened 2022-11-18
[feature] use potrf from cuSolver

toxa81:
Mystery: it works in a standalone test, but not inside wf::orthogonalize(). Invetigation is in progress 🧐


toxa81:
Summary: this are the preparatory modifications to start using potrf() from cuSolver. However, currently it is not working for unknown reason and thus commented :)

PR #784, opened 2022-11-19
[fix] Allocate wave-function from memory pool

toxa81:
Small fix that improves a lot allocation of the host-pinned memory for the wave-functions. I tested it on Au-surf example and the performance is good now.

As a by-product: use std::make_unique in a couple of places and allow to set parallel FFT transformation for the wave functions if mpi grid is of the [n, 1] shape.


toxa81:
Is is OK to merge it?

simonpintarelli (approved):
lgtm

PR #796, opened 2022-12-19
Update master branch

toxa81:
@mtaillefumier this is what I mean. Why in hell there are so many conflicts if `master` is only a retarded version of `develop` and we never commit diverging (or any) changes to master directly. I'm puzzled. 🧐🤔


mtaillefumier:
The last commit dates back from Oct. 5th in master. I also have this commit locally. So anything before that should not be showing.

toxa81:
Exactly! I think the issue is with squash commits in develop, but I can't prove it. I started a new `main` branch based on current develop. Should be good enough. I will try to keep `main` updated more frequently. It all goes well, we will keep it and remove `master`.

PR #798, opened 2022-12-19
[ELPA] Fit latest elpa into the code

toxa81:
Several issues with the latest ELPA were addressed:
* elpa now defines it's own `double_complex` and `float_complex` macros. To avoid the name collistion, the decision is taken to remove all our `using double_complex = std::complex<double>` aliases
* linking problem is reported here: https://github.com/marekandreas/elpa/issues/30
* our `elpa.hpp` header was removed as unnecessary


mtaillefumier (approved):
Do we bump a new version for sirius ?

toxa81:
Anything after 7.4 would be ok in terms of API. We can make the minor release any time before the CP2K release.

PR #803, opened 2022-12-26
[enhancement] Augmentation operator on GPUs

toxa81:
TODO:
 - [x] forces
 - [x] stress
 - [x] charge density augmentation
 - [x] D-matrix construction
 - [x] add control input parameter for the G-vector chunk size
 - [x] ~generate Rlm for a block of G-vectors, then use for all atom types~
 - [x] use streams to compute augmentation for density and magnetisation in parallel
 - [x] use streams to compute D-matrix in parallel

With this changes it is now possible to run the large cutoff simulations.


toxa81:
ci run

toxa81:
ci try

toxa81:
Modify the code to run large plane-wave cutoffs. In this case even local number of G-vectors has to be split in chunks such that augmentation operator can fit in the GPU memory. 
* augmentation operator was simplified; prepare() and dismiss() functions were removed
* augmentation operator derivative class was incorporated in the  Augmentation_operator class

simonpintarelli (commented):
lgtm

PR #807, opened 2023-01-11
[refactor] linear algebra la:: namespace

toxa81:
* move BLACS_grid, dmatrix, Eigensolver under the la:: namespace
* `linalg_cost<T>` is renamed to `constant<T>`
* `linalg_t` is renamed to `lib_t`
* `linalg` class is renamed to `wrap`

For example, now the call to gemm() looks the following: `la::wrap(la::lib_t::blas).gemm(...)`

The code logic is unchanged.


mtaillefumier (approved):
the majority of the modification is renaming namespace etc... No change in the logic of code just cleaning it.

suggestion : squash merge all commits.

PR #813, opened 2023-01-25
[fix] Use MPI grid for setting fft communicators

toxa81:
This PR isolates the work on setting the parallel FFT communiator using MPI grid from the PR (https://github.com/electronic-structure/SIRIUS/pull/803). Now the MPI grid for the band parallelisation is used to setup parallel FFT communcators.


toxa81:
cscs-ci run default

PR #814, opened 2023-01-30
[fix] Change spglib tolerance from 1e-4 to 1e-6

toxa81:
Use smaller value of tolerance parameter to detect symmetries of crystall and lattice.


toxa81:
@mtaillefumier let's check this one and then I can make a new release.

toxa81:
cscs-ci run default

mtaillefumier (approved):
good to go. It is probably the best approach we can find without fixing spglib

PR #820, opened 2023-02-15
[enhancement] Simplify the implementation of non-local stress.

toxa81:
Also in this PR:
- pass gk_cutoff as a command-line parameter to mini-app
- introduce dRlm_dr_numerical() for debug purposes


mtaillefumier (approved):

mtaillefumier:
let check the formula just to avoid a simple mistake

mtaillefumier (approved):
looks alright

PR #827, opened 2023-03-21
[enhancement] introduce Spheric_function_set class

toxa81:
Introduce a set of spherical functions and symmetrization of the set. The first use case  is the PAW Vxc and Exc functions. Now the potential is computed first, then symmetrized and only then the contribution to total energy is computed. As by-product, the PAW code itself is cleaned up and simplified a little bit.


toxa81:
@simonpintarelli That's the main lines of the PR https://github.com/electronic-structure/SIRIUS/blob/enhancement/spheric-function-set-v3/src/potential/paw_potential.cpp#L81-L94

I hope this can improve the NLCG convergence of PAW. I have one more idea to try afterwards.

PR #828, opened 2023-03-22
[feature] Umpire memory pool

mtaillefumier:
- Use umpire for the memory pool (supports both CPU and GPU memory)
- Updated the local spack recipe

TODO:
- [x] @toxa81 upload latest Dockerfile for github
- [x] @toxa81 test simplification of SPEC with just using `^umpire+cuda~device_alloc` (doesn't work)
- [x] @mtaillefumier fix the Umpire exception in test
- [x] @mtaillefumier clean up all intermediate patches and fixes introduced to understand the linking issue
- [x] @mtaillefumier build github ci/cd image for GPU; it should run on CPUs just fine 
- [x] @mtaillefumier update sirius recipe on spack repo
  - [x] @toxa81 trigger rebuild of base images for K8S build farm
- [x] get rid of DEBUG_MEMORY_POOL


toxa81:
@mtaillefumier what is the plan for spack recipe? To have "+memory_pool" and "+umpire" and then get rid of "+memory_pool"?

mtaillefumier:
https://github.com/toxa81 : we can not get rid off of the memory_pool option because someone may want to compile a specific (outdated version of sirius) that does not have umpire support. 
The idea is to keep memory_pool and compile umpire if memory_pool is turned on. SIRIUS supporting it will pick it up and the older versions will simply ignore it. The recipe will need a second update after we have the first official version SIRIUS with umpire support.
I see no other option the time being as we rely on the official recipe in the CI.

simonpintarelli:
Why not just use `+memory_pool` also for the umpire enabled version and completely remove the internal memory pool implementation from now on?

We should use a local spack repo to build the base image, sending a PR first to spack upstream seems not right.

mtaillefumier:
cscs-ci run default

mtaillefumier:
we need to add this 

RUN spack repo add /sirius-src/spack

in the ci otherwise it will always fail.

toxa81:
>RUN spack repo add /sirius-src/spack
It's added in the base container. The f**er doesn't recongnize the installed packages. All works fine on my workstation.

toxa81:
@simonpintarelli The issue is this
```

Run spack find -l
  spack find -l
  shell: sh -e {0}
  env:
    DEVSPEC: sirius@develop %gcc +python +tests +apps +vdwxc +scalapack +fortran build_type=RelWithDebInfo ^openblas ^mpich arch=linux-ubuntu22.04-x86_64
    ```
    
  `spack find` inside container on github shows nothing

**Comment by toxa81** [comment]
@mtaillefumier I fixed the ci/cd container. Now umpire is there and linking fails `__cudaRegisterLinkedBinary_fac9bec3_19_DeviceAllocator_cpp_54a80de0'

**Comment by mtaillefumier** [comment]
thanks. I will look at this tomorrow

**Comment by mtaillefumier** [comment]
I can not reproduce this issue locally so I am being pedantic for the time being.

**Comment by mtaillefumier** [comment]
we keep having this warning
`
cc1plus: warning: command-line option '-fallow-argument-mismatch' is valid for Fortran but not for C++`

**Comment by simonpintarelli** [comment]
> we keep having this warning ` cc1plus: warning: command-line option '-fallow-argument-mismatch' is valid for Fortran but not for C++`

Something is adding this is in the docker container. It breaks rocm compilation at the moment, but I haven't been able to figure out who/what is adding that flag to CXX.

**Comment by toxa81** [comment]
@mtaillefumier 
* the modifications to umpire package are not seen by the base container
* why do you need to modify umpire? it's working for others w/o patches. or not?

**Comment by mtaillefumier** [comment]
because umpire is still configured with `^umpire@2022.03.1%gcc@11.3.0+c+cuda+device_alloc`. I want the `+device_alloc` gone for the time being as I think it is the root of the problem. It is turned off by default in their cmake configuration and I have no issue with sirius+umpire if I leave the default cmake options for umpire.

**Comment by msimberg** [comment]
@mtaillefumier just FYI, turning off `device_alloc` is perfectly reasonable. If you don't want to make a hard requirement to have that off you should also be able to enable separable compilation (like https://github.com/aurianer/DLA-Future/blob/367cf3517510d113e2df060dec425e15b4d53dd8/src/CMakeLists.txt#L288). According to this comment they're thinking about providing a separate library for the device side functionality which would allow "normal" linking without any extra work, and if one links to the device functionality one may have to take additional steps.

**Comment by msimberg** [comment]
> Why not just use `+memory_pool` also for the umpire enabled version and completely remove the internal memory pool implementation from now on?
> 
> We should use a local spack repo to build the base image, sending a PR first to spack upstream seems not right.

That's one option. Another would be to simply point CI to a PR (or not PR) branch with the required change until it has made it upstream.

File: CMakeLists.txt [hunk 1]
```diff
@@ -138,6 +138,10 @@ include(cmake/hdf5_target.cmake)
 find_package(SpFFT 1.0.2 CONFIG REQUIRED)
 find_package(SPLA 1.4.0 CONFIG REQUIRED)
 
+if (USE_MEMORY_POOL)
```
msimberg, line 141: Sorry for the drive-by comment (as I'm sure this applies to other options as well): *please* namespace your CMake variables.

toxa81: That is a big TODO :)

mtaillefumier:
__cudaRegisterLinkedBinary_fac9bec3_19_DeviceAllocator_cpp_54a80de0 is the problem I trying to solve. We are not using umpire allocators on the device side as all device memory is allocated directly by the host (basically no malloc in a gpu function). It is turned off by default in umpire (master from this week or so) that's why I am really puzzled why they turned it on by default in their spack recipe.

src/CMakeLists.txt
```diff
@@ -89,7 +89,7 @@ if(USE_ROCM)
 endif()
 
 if(USE_CUDA)
-  set_target_properties(sirius PROPERTIES CUDA_ARCHITECTURES ${CMAKE_CUDA_ARCHITECTURES})
+  set_target_properties(sirius PROPERTIES CUDA_ARCHITECTURES ${CMAKE_CUDA_ARCHITECTURES} CUDA_RESOLVE_DEVICE_SYMBOLS ON)
```
toxa81, line 92: Wo we even need to do this line? According to CMkae doc. CMAKE_CUDA_ARCHITECTURES provides a default for CUDA_ARCHITECTURES.

mtaillefumier: CUDA_ARCHITECTURES can be removed that's not the issue.

toxa81:
@mtaillefumier is there something I should do with the base image? How it works for other projects?

toxa81:
@msimberg  why DLA-F links to Umpire without issues?

simonpintarelli:
I've tried it locally. It links with `~device_alloc`. The only culprit is that `sirius~cuda` conflicts with `umpire+cuda`. Therefore the baseimage needs to be rebuilt in order to provide `umpire+cuda` and `umpire~cuda`.

mtaillefumier:
There is nothing wrong with umpire. The issue is that we use packages compiled with GPU support even when we want to compile SIRIUS in CPU mode only. It does not seem to matter much for spla and spfft but it does matter (a lot) for umpire.
The solution to this problem is to compile umpire in cpu mode only or to have two versions of umpire one with cpu mode and one with cpu+gpu mode.

simonpintarelli:
umpire in the baseimage is present with `+device_alloc`.

https://github.com/electronic-structure/SIRIUS/actions/runs/4598307014/jobs/8122165932?pr=828#step:4:83


We need to check what is the best option to include the 2nd version of umpire.

simonpintarelli:
> then give me the two exact specs that are needed. My guess: umpire%gcc+cuda umpire%gcc~cuda

It is now also reflected in the spack recipe:
https://github.com/electronic-structure/SIRIUS/blob/umpire/spack/packages/sirius/package.py#L247

Adding two different specs to the container would work as well.

Or one spec per config, e.g. for each item in the list
https://github.com/electronic-structure/SIRIUS/blob/umpire/.github/workflows/build.yml#L9

Not sure this is needed though, it looks like the manual calls to `cmake` with the `USE_XXX` flags worked too.

toxa81:
last Q: do we need `+device_alloc`?

simonpintarelli:
> last Q: do we need `+device_alloc`?

`+device_alloc` is not needed (see Mikaels comment)

toxa81:
cscs-ci run default

toxa81:
Ok! We wait for K8S build farm CI/CD

mtaillefumier:
cscs-ci run default

mtaillefumier:
spack recipe pr https://github.com/spack/spack/pull/36958/files

toxa81:
@mtaillefumier here comes trouble `ld.lld: error: unable to find library -lumpire`

mtaillefumier:
we have to use the local spack recipe included in sirius to build the images.

electronic-structure:
I didn't use it fir cuda images, but it works. The upstreamed spack recipe is correct.

simonpintarelli:
The following line adds only `-lumpire`:
https://github.com/electronic-structure/SIRIUS/blob/umpire/src/CMakeLists.txt#L114

It looks like the  `find_package(umpire ...)` in the current cmake.

toxa81:
Do you know what is wrong with rocm build?

mtaillefumier:
> The following line adds only `-lumpire`: https://github.com/electronic-structure/SIRIUS/blob/umpire/src/CMakeLists.txt#L114
> 
> It looks like the `find_package(umpire ...)` in the current cmake.

that's also correct.

simonpintarelli:
> > The following line adds only `-lumpire`: https://github.com/electronic-structure/SIRIUS/blob/umpire/src/CMakeLists.txt#L114
> > It looks like the `find_package(umpire ...)` in the current cmake.
> 
> that's also correct. umpire has no namespacing at all

Indeed the namespacing is missing, quite malicious.

PR #830, opened 2023-03-31
[enhancement] Periodic_function class

toxa81:
TODO:
 - [x] use Spheric_function_set in Periodic_function
 - [x] collect pointers from Fortran in a data structure
 - [x] set pointers to Fortran function when Periodic_function is created
 - [x] @simonpintarelli adjust Python interface, especially for strong type lmax_t
 - [x] @toxa81 test parallel runs
 - [x] @toxa81 fix and test Exciting code
 
Summary of the PR:
 - use overloaded `copy()` function to copy opbjects; mdarray uses the same; operator>> is removed
 - Periodic_function is now a composition of two objects: Smooth_periodic_function (defined on the FFT grid) and Spheric_function_set (defined for the muffin-tin part)
 - Wrappers for external pointers to MT and RG parts of the Periodic_function are introduced
 - API functions to set and get Periodic_function are changed to handle the dimensions of the functions correctly
 - strong type lmax_t is introduced
 -  SDDK/type_definition.hpp header is removed
 - strong_type.hpp header is introduced
 - unused commented code in symmetrize.hpp is removed
 - Density::magnetisation() renamed to Density::mag() to be consistent with Density::rho()
 - use auto keyword in more places
 - copy(), scale() and axpy() are introduced for mixing Periodic_functions
 - minor clean-ups


toxa81:
cscs-ci run default

toxa81:
@mtaillefumier sorry for the monstrous PR

python_module/py_sirius.cpp
mtaillefumier, line 1: unless using the old commented code is planned, I would suggest to delete it completely. We can always recover it with git

mtaillefumier (approved):
everything looks alright. We can squash all commit in one when merging.

electronic-structure:
squash is the only available option, I will do the merge

python_module/py_sirius.cpp
simonpintarelli: I removed all commented code in `py_sirius.cpp`, although it's not related to this PR.

PR #831, opened 2023-04-24
Add missing ifdef for umpire

simonpintarelli:
```
spack install sirius~memory_pool
``` 
failed to compile because a missing ifdef in `memory_pool::clear`.

Not related, but `struct Real` has been defined twice. I've included the header `typedefs.hpp` in `acc_common.hpp` and removed the duplicate.


toxa81 (approved):
LGTM.

PR #832, opened 2023-04-25
[cmake] Use SIRIUS namespace convention for options and other SIRIUS …

mtaillefumier:
- [x] Namespace cmake options.
- [x] Namespace the SIRIUS compilation flags
- [x] The spack recipe is modified to reflect the cmake changes.
- [x] Use the local spack recipe for sirius instead of the official version (We need to update it eventually)
- [x] updated the docker containers for the CI.
- [x] Update the docker containers for app distribution (we should do nothing here)
- [x] Update the spack recipe in the spack repo. Need a PR for this.


simonpintarelli:
Do we need the namespaces in the options? Isn't this only an advantage if the project is included as subdirectory?
I'd rather have the fortran api with improved linking, e.g. almost everything internal can be made private, which means we could almost delete `siriusConfig.cmake`.

simonpintarelli:
I agree it seems the prefixed option names is common.  But what is the benefit for SIRIUS? As far as I can see the options `USE_XXX` etc don't show up in the installation directory `lib/cmake`. 

For the Fortran API I don't mean to retire cmake config, just correctly declaring the dependencies, at the moment anything is public.

mtaillefumier:
USE_OPTION can be used by other projects as well which is also common. USE_CUDA for instance is a typical example of this issue and it can have disastrous consequences as we encountered in cosma / costa for instance. I had similar issues when working on cp2k cmake build system. It is unfortunate that cmake does not automatically do this so we have to compensate for this design flaw.

mtaillefumier:
it is a bad idea indeed to build sirius in tree (that's why submodule is also a bad idea). The problem is identical to the DFLAGS in header files provided by different libraries. `-DUSE_MKL` might mean something different for dependency A and B and might trigger branches in their respective header files that might only be supported if the dependencies A and/or B are compiled with the relevant flags. That's why for instance compilation dependent flags should avoid __NAME because __NAME is usually reserved to compilers. 
Namespacing

mtaillefumier:
> I agree it seems the prefixed option names is common. But what is the benefit for SIRIUS? As far as I can see the options `USE_XXX` etc don't show up in the installation directory `lib/cmake`. In `siriusConfig.cmake` the options are substituted with the value before installation, therefore it's unclear to me what is the benefit of the namespace.
> 

well we do things properly but other do not. ;)


> For the Fortran API I don't mean to retire cmake config, just correctly declaring the dependencies, at the moment anything is public.

I misunderstood your intend here for the fortran api. I though you were referring to the problem you mentioned yesterday. 

Indeed everything is public in cmake and I do not know how to make things private by default. I am not even sure it can be done actually.

simonpintarelli (approved):
lgtm

PR #836, opened 2023-05-12
[fix] Umpire allocator for the multiple device case

toxa81:


toxa81:
@mtaillefumier 
```
getAllocationStrategy Allocator "DEVICE0" not found. Available allocators: 
```

mtaillefumier:
two semi-column `::` are missing. It should be "DEVICE::0" not DEVICE0

mtaillefumier:
You were faster than me.

toxa81:
I confirm the fix works. I can now run again on bare metal hohgant.

mtaillefumier (approved):

PR #839, opened 2023-05-31
[enhancement] Minor updates

toxa81:
* minor coding style fixes
* avoid referencing the GPU memory pool when running on a host without device; umpire is unhappy about this


simonpintarelli (approved):
Thanks. LGTM. `using namespace sddk` in a `.cpp` file should be ok, but not in a header file.

simonpintarelli (approved):
Thanks. LGTM. `using namespace sddk` in a `.cpp` file should be ok, but not in a header file.

PR #840, opened 2023-06-01
[enhancement] refactor radial function and basis indices

toxa81:
Summary of the PR
* angular_momentum class is introduced to handle pure orbital momentum and full orbital momentum
* new implementation of radial_function_index class that supports iterator 
* new implementation of basis_function_index class that supports iterator
* strong types for radial and basis indices are introduced
* obscure variable offset_lo is removed both from atom_type and atom classes
* beta projectors are no longer treated as local orbitals; local orbitals are now only used in LAPW context
* double total_angular_momentum is removed from local_orbital_descriptor struct and replaced by angular_momentum instance
* some variables in the Unit_cell are now computed on the fly, without precomputing and storing them as the class members


toxa81:
cscs-ci run default

mtaillefumier (approved):

PR #846, opened 2023-06-14
[update] hdf5 i/o

toxa81:
Add fortran API to save and load state file.


simonpintarelli (approved):
LGTM. Might be a candidate for future refactoring, e.g. for example to pass the filename from the fortran api, such that it can be used together with vc-relax (not sure this is needed tough).

toxa81:
> LGTM. Might be a candidate for future refactoring, e.g. for example to pass the filename from the fortran api, such that it can be used together with vc-relax (not sure this is needed tough).

good point! I will add it to this PR

PR #847, opened 2023-06-15
remove AUR package from readme

simonpintarelli:
Remove archlinux AUR from readme. The aur package isn't up-to-date, and I think spack should be used. Apparently it had no users, otherwise someone would have complained about missing spla dependency...


toxa81 (approved):
LGTM

PR #850, opened 2023-06-25
[feature] New split index implementation

toxa81:
Split index class is simplified. It is now teamplated over the index type which allows to distingush between different indices. Iterator is introduced:
```
for (auto it : spl_index) {
  std::cout << "local index : " << it.li << ", global index : " << li.i << std::endl;
}
```
Iterator has two fields: `i` for global index and `li` for local index.

### DIFF
--- a/apps/tests/test_allgather.cpp
+++ b/apps/tests/test_allgather.cpp
@@ -7,10 +7,10 @@ void test_allgather()
     int N = 11;
     std::vector<double> vec(N, 0.0);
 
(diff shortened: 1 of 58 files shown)


simonpintarelli:
When iterating over kpoints, it would nice if one could do the following:

```
for(auto& it : kpoint_set) {
   auto& kpoint = *it;
   it.local_index;
   it.global_index;
}
```
Similar for atoms in the unit_cell. Not sure it's a good idea though, e.g. if it fits for the other quantities or would require a separate implementation.

toxa81:
replaced by https://github.com/electronic-structure/SIRIUS/pull/862

PR #852, opened 2023-06-29
[enhancement] Simplify Simulation_context class.

toxa81:
Collect all radial integrals and all callback functions in standalone data-structures. Rationale: they are internal objects that do not need to be properly "interfaced". This simplifies the amount of methods and members of Simulation_context class.


simonpintarelli (approved):
lgtm

PR #861, opened 2023-07-12
Feature/split index proposal

simonpintarelli:
Proposal for changes in #850:

Change `operator*` to:

```
    inline auto operator*()
    {
        return std::make_pair(idx_->global_index(this->li), this->li);
    }
```

And then use `for (auto [i, li] : spl_index) { }`. Structured bindings is a c++17 feature, but it seems to work when using c++14 standard too (adding `-Wno-c++17-extensions` to suppress noise)


toxa81:
I'm working in rebase, should be ready by tomorrow

simonpintarelli:
Unfortunately the structured bindings don't work together with openmp/clang: https://godbolt.org/z/GaM374fe4

Locally clang is even crashing when the index is used inside the openmp region, which occurs several times.

PR #862, opened 2023-07-12
[feature] new split index implementation

toxa81:


simonpintarelli:
I've changed `operator*` to return an unnamed struct containing `Index::global i, Index::local li`, and removed the global index from the class variables. `i` was not updated when calling `operator++`, and without knowing how the class is implemented, one might use the following code:

```
for(auto it = begin(spl_index); it != end(spl_index); ++it) {
   (void) it.i; // Global index is accessible, but not updated
   (void) *(it.i).i; // Global index is set only in operator*
}
```

simonpintarelli (approved):
lgtm

PR #864, opened 2023-07-13
[feature] implement SIRIUS_CONFIG environment variable

toxa81:


mtaillefumier (approved):
I have no objection about this but I do not really understand why it is better to pass the file using an env variable instead of the old behavior.
The only use case I could think off is aiida.

toxa81:
> I have no objection about this but I do not really understand why it is better to pass the file using an env variable instead of the old behavior. The only use case I could think off is aiida.

There is also Exciting where implementing new command line argument is not encouraged. But some times I need to pass modifications to the options.

PR #884, opened 2023-08-21
[feature] restore function to set density matrix

toxa81:
We need to set density matrix to compute D_{ij} of PAW, otherwise the Hamiltonian definition is incomplete.


simonpintarelli (approved):
LGTM

PR #890, opened 2023-08-28
use const where possible

simonpintarelli:
WIP (after #889).

- Modify all functions calling `H_k.apply*` to take `Hamiltonian_k` as const reference and a separate argument for non-const ref to `K_point`.
- Some other minor fixes proposed by `-Wextra` flag.
- add `Hamiltonian_k::apply_S` for the overlap operator (for LR)


toxa81:
If we now bring K_point as separate object, it must probably be removed as a member of Hk to avoid confusion.

simonpintarelli:
> If we now bring K_point as separate object, it must probably be removed as a member of Hk to avoid confusion.

Thanks, I've now removed the public member function to access the kpoint from Hk.

PR #896, opened 2023-09-28
[refactor] better organise namespace

toxa81:
* TERMINATE() macro is removed; use RTE_THROW instead
* STOP() macro removed; use RTE_THROW instead
* WARNING() macro is removed; use RTE_WARNING
* in many paces assert() macro is replaced by RTE_ASSERT()
* utils:: namespace is removed
  - math functions are moved to math_tools.hpp
  - time functions are moved to time_tools.hpp
  - output stream functions are moved to ostream_tools.hpp
  - string functions are moved to string_tools.hpp
* some files are moved to ./core folder in preparation to remove SDDK
* splindex class is moved out of sddk:: to sirius:: namespace
* old interface to Plasma eigensolver is removed
* some unused gpu kernels are removed
* acc:: namespace is re-organised and moved to ./core

This PR touches a lot of files, but it doesn't change the logic of the code. Only source file reshuffling and namespace organisation. Unused cuda code is removed. What is left is la:: space, but this will be done in separate PR after https://github.com/electronic-structure/SIRIUS/pull/900


toxa81:
cscs-ci run default

simonpintarelli (approved):
LGTM

apps/mini_app/sirius.scf.cpp
```diff
@@ -412,34 +412,34 @@ void run_tasks(cmd_args const& args)
...
-        Simulation_context ctx2(dict1["config"].dump());
-        ctx2.initialize();
-
-        ctx1.unit_cell().set_lattice_vectors(ctx2.unit_cell().lattice_vectors());
-        for (int ia = 0; ia < ctx2.unit_cell().num_atoms(); ia++) {
-            ctx1.unit_cell().atom(ia).set_position(ctx2.unit_cell().atom(ia).position());
-        }
-        ctx1.update();
-    }
+    //if (task_id == task_t::read_config) {
```
mtaillefumier, line 415: remove dead code.

mtaillefumier (approved):
I wonder if we can just put ETH Zurich / CSCS instead of the authors because we can always trace the authors back with github and commits.

simonpintarelli:
> I wonder if we can just put ETH Zurich / CSCS instead of the authors because we can always trace the authors back with github and commits.

The git blame is not necessarily the same as the author, due to reformatting or other cosmetic changes.

toxa81:
1. I saw in other codes that each file is equipped with boile plate
2. I will be happy to switch to ETH Zurich / CSCS ownership or even sirius-org. I just need to find out if I can "just do it".

mtaillefumier:
> > I wonder if we can just put ETH Zurich / CSCS instead of the authors because we can always trace the authors back with github and commits.
> 
> The git blame is not necessarily the same as the author, due to reformatting or other cosmetic changes.It does not scale with projects that have far too many devs. We are three so it is ok.

mtaillefumier:
we do not have to do it now, I wanted to put the question on the table because we will eventually have to do something like this. I guess we can just leave it as is until SIRIUS really see external developers contribute to it.

mtaillefumier:
Licencing each file is also fine.

PR #898, opened 2023-10-11
Fix ScaLAPACK p*syevd

RMeli:
While working on DLA-Future integration in SIRIUS, I stumbled upon what looks like a spurious recursion. Fixing it also uncovered some missing `reinterpret_cast` in `solve_` (for the eigenvalues and the workspace).

### DIFF
--- a/src/linalg/eigenproblem.hpp
+++ b/src/linalg/eigenproblem.hpp
@@ -614,13 +614,17 @@ class Eigensolver_scalapack : public Eigensolver
         if (std::is_same<T, double>::value) {
             FORTRAN(pdsyevd)
             ("V", "U", &matrix_size__, reinterpret_cast<double*>(A__.at(sddk::memory_t::host)), &ione, &ione,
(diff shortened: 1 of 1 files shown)


simonpintarelli:
Thanks @RMeli . Afaik the full scalapack eigensolver is not used in the code, only the routines which find n-lowest eigenvalues. 

I don't like the `reinterpret_cast` in the `if(std::is_same<T, X>::value)`, we can now actually use `constexpr if`. I'll try to fix it in a separate PR.

simonpintarelli (approved):
LGTM

RMeli:
> the full scalapack eigensolver is not used in the code

If it's not used another approach would be to remove it?

> I don't like the `reinterpret_cast` in the `if(std::is_same<T, X>::value)`, we can now actually use constexpr if.

Good point. Now that we use C++17 we could also use the less verbose `std::is_same_v` instead.

mtaillefumier (approved):
looks alright

PR #902, opened 2023-10-12
for compatibility to gcc-13

yizeyi18:
Added #include <cstdint> to certain files for compatibility to gcc-13.

### DIFF
--- a/src/SDDK/memory.hpp
+++ b/src/SDDK/memory.hpp
@@ -29,6 +29,7 @@
 #include <iostream>
 #include <map>
 #include <memory>
+#include <cstdint>
 #include <cstring>
 #include <functional>
 #include <algorithm>
--- a/src/typedefs.hpp
+++ b/src/typedefs.hpp
@@ -26,6 +26,7 @@
 #define __TYPEDEFS_HPP__
 
 #include <cstdlib>
+#include <cstdint>
 //#include <assert.h>
 #include <complex>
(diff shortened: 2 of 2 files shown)


toxa81:
cscs-ci run default

toxa81:
Dear @yizeyi18 
sorry for late repy. Can you plese do the same PR, but base it on `develop` branch.

yizeyi18:
Dear@toxa81
Sorry, I am new to github and don't know how to change its branch. Can you help to move it?
This change is just for compatibility with gcc-13 and I missubmitted it to master branch. Sorry.

simonpintarelli:
The develop branch doesn't need modifications, cstdint includes have been added before, it compiles with gcc13.

yizeyi18:
> The develop branch doesn't need modifications, cstdint includes have been added before, it compiles with gcc13.

seems src/SDDK/memory.hpp had not added yet. Does it really compiles with gcc-13?
I only tried release-7.4.3 and it does not.

simonpintarelli:
@yizeyi18 Yes, including cstdint in typedefs.hpp is enough to make it compile. 

The target branch can be changed using the `Edit` button next to the title. You would have to rebase your commits to the develop branch first.

PS: In order to be correct, cstdint would have to be included in every file where `unitX_t ` is referenced, these would be a few more according to grep.

PR #904, opened 2023-10-17
sync package.py with upstream spack

simonpintarelli:
Sync local package.py with spack@develop, verified using diff with spack@develop (where boost is still present):

```
82,88d81
<     variant(
<         "boost_filesystem",
<         default=False,
<         description="Use Boost filesystem for self-consistent field method "
<         "mini-app. Only required when the compiler does not "
<         "support std::experimental::filesystem nor std::filesystem",
<     )
110d102
< 
162d153
<     conflicts("+boost_filesystem", when="~apps")
```


spack/packages/sirius/package.py
```diff
@@ -263,18 +241,16 @@ def cmake_args(self):
             args.append(self.define(cm_label + "ELPA_INCLUDE_DIR", elpa_incdir))
 
         if "+cuda" in spec:
-            cuda_arch = self.spec.variants["cuda_arch"].value
+            cuda_arch = spec.variants["cuda_arch"].value
             if cuda_arch[0] != "none":
-                args += [self.define("CMAKE_CUDA_ARCHITECTURES", cuda_arch)]
+                # Make SIRIUS handle it
+                if "@6:7.4.3" in spec:
```
toxa81, line 247: didn't we retire <7 versions of sirius?

simonpintarelli: Yes, the lower bound is actually not neede here, I made it `:7.4.3` now.

PR #906, opened 2023-10-24
[update] Dockerfile for CI

toxa81:


simonpintarelli (approved):
Why is ldconfig required? 

LGTM

toxa81:
>Why is ldconfig required?
It's the way how Sarus MPI hook is informed.

PR #913, opened 2023-11-04
[refactor] get rid of sddk:: namespace and update mdarray class

toxa81:
Refactor mdarray class and get rid of sddk:: namespace


mtaillefumier (approved):

simonpintarelli (approved):
lgtm

PR #918, opened 2023-11-07
[CI] Enable format check

toxa81:
* change the python format scipt
* clang-format >=17 is required
* ci/cd workflow to check format is enabled
* old scripts are removed
* apply_format.x script is introduced


toxa81:
@mtaillefumier it's ok that the source code formatting doesn't pass. It will be checked in the real "formatting" PR.

mtaillefumier (approved):

PR #920, opened 2023-11-08
[format] apply clang-format to all files

toxa81:


toxa81:
@simonpintarelli this looks good now. The plan is to squash merge it and introduce `.git-blame-ignore-revs` in the second PR for the squashed commit

mtaillefumier:
one thing. `apply-format.x` should actually format the file not return the diff. it is not practical to save the diff and then apply it manually.

toxa81:
> one thing. `apply-format.x` should actually format the file not return the diff. it is not practical to save the diff and then apply it manually.

the script is doing exactly this; you need to load clang-format 17.0.4

PR #922, opened 2023-11-15
[refactor] Beta projectors

toxa81:
* optimisations to create and re-use beta projectors
  * pw coefficients of beta-projectors for atom types are always stored on the device
  * device storage for all coefficients for all atoms is controlled by a switch
  * beta_chunk descriptor is not copied; instead a pointer is used
  
 TODO @simonpintarelli for some reason .def("make_generator", .. in the Python module started to complain. I commented it out, but it has to be restored. Can you please have a look?


toxa81:
cscs-ci run default

simonpintarelli (approved):
LGTM. Not sure what happened to the format check, it seems to fail in unrelated files.

electronic-structure:
format is not applied yet because of these last minute PRs.

PR #923, opened 2023-11-16
Pin DLA-Future version in Spack package with some fixes and improvements

RMeli:
* Pin DLA-Future to freshly released version `0.3.0`

Improvements
* Enable `dlaf` variant only with `@develop`
* Change repo name from `sirius` to `sirius-repo` to avoid confusion with package name

Fixes:
* Add `SIRIUS_USE_DLAF` so that `spack install sirius+dlaf` actually works
  * Previously I only used the package to manage dependencies, so I missed this
* Remove `SIRIUS_` prefix from `BUILD_SHARED_LIBS`

`pika+cuda` sets `INTERFACE_POSITION_INDEPENDENT_CODE=ON`, therefore it is incompatible with the hard-coded `PISITION_INDEPENEDENT_CODE=OFF` in some of the apps/utils. I removed this constraint, but I'm not sure why it was there in the first place.

### DIFF
--- a/apps/mini_app/CMakeLists.txt
+++ b/apps/mini_app/CMakeLists.txt
@@ -7,7 +7,6 @@ if(SIRIUS_USE_VCSQNM)
   target_link_libraries (sirius.scf PRIVATE Eigen3::Eigen sirius)
 endif()
 install(TARGETS sirius.scf RUNTIME DESTINATION "${CMAKE_INSTALL_PREFIX}/bin")
-set_property(TARGET sirius.scf PROPERTY POSITION_INDEPENDENT_CODE OFF)
 
 # Register verification tests as well
 set(SIRIUS_SCF_LABELS cpu_serial cpu_band_parallel)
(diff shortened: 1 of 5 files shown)


mtaillefumier:
I would leave the POSITION_INDEPENDENT_CODE=ON. If applications still rely on POSITION_INDEPENDENT_CODE=OFF then they should be fixed.... There is no penalty in using POSITION_INDEPENDENT_CODE=ON by default with both static (.a) and dynamic libraries (.so)

RMeli:
> I would put POSITION_INDEPENDENT_CODE=ON

@mtaillefumier, I think this is indeed the case (implicitly)?
https://github.com/electronic-structure/SIRIUS/blob/76cc9ab0e8f9f108d8a8966cd8a4e3cf020f468a/CMakeLists.txt#L57

mtaillefumier:
then all good.

simonpintarelli (approved):
LGTM. I think the `POSITION_INDEPENDENT_CODE OFF` dates back to the python interface. The python library required position independent code,  I don't remember why it was turned off explicitly everywhere else.

toxa81 (approved):
LGTM

PR #927, opened 2023-11-21
[api] add two options to lr solver

toxa81:
The change to API is not breaking because the parameters are optional.

### DIFF
--- a/src/api/sirius_api.cpp
+++ b/src/api/sirius_api.cpp
@@ -6042,6 +6042,14 @@ sirius_linear_solver:
       type: int
       attr: in, required
       doc: Number of occupied bands.
+    tol:
+      type: double
+      attr: in, optional
+      doc: Tolerance for the unconverged residuals (residual L2-norm should be below this value).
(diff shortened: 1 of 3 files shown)


mtaillefumier (approved):

src/api/sirius_api.cpp
```diff
@@ -6193,11 +6202,15 @@ sirius_linear_solver(void* const* handler__, double const* vkq__, int const* num
 
                 // Identity_preconditioner preconditioner{static_cast<size_t>(nbnd_occ)};
 
+                auto tol = get_value(tol__, 1e-13);
+
                 auto result = sirius::cg::multi_cg(linear_operator, preconditioner, X_wrap, B_wrap, U_wrap,
                                                    C_wrap, // state vectors
                                                    100,    // iters
```
gsavva, line 6209: probably we have to remove the hard-coded max number of iterations as well, but that can be done in another PR

gsavva (approved):

PR #928, opened 2023-11-22
[hubbard] Use enum type to set up the method for generating the hubbard subspace

mtaillefumier:
content of the  PR 
- regroups three different input parameters controlling how the hubbard subspace is generated in to one parameter only. The API already offers this to some extend.
- Adding atomic orbitals for the hubbard subspace through the Fortran interface assumes that the atomic type is fully initialized which is not the case when the potential information are stored in external files.


src/context/DOC.md
toxa81, line 1: This file should not be added, it is a by-product of the api generator script and still WIP

toxa81 (changes requested):
DOC.md needs to be removed, otherwise it's fine

toxa81:
> We may need a new version of sirius before I can open a PR on cp2k side of things

I fully agree, we are ready to tag 7.5

src/context/DOC.md
mtaillefumier: Ok I will fix it

mtaillefumier: removed.

PR #931, opened 2023-11-30
[fix] add .template keyword

toxa81:
for some compilers .template is needed to access radial_integrals_sum<>()


simonpintarelli:
Why not pass the enum as argument instead of template parameter?

PR #934, opened 2023-12-05
[fix] subspace initialization in test_lr_solver

toxa81:
It should be complex, i.e. `std::complex<F>`, not real.


gsavva (approved):
I've had instances where this was not failing locally...

PR #951, opened 2024-01-04
[fix] use RTE_THROW() macro in more places

toxa81:
I'm debugging cp2k+sirius which fails in strange places. First thing to do is to have meaningful error messages in the places where it fails.


simonpintarelli (approved):
lgtm

PR #954, opened 2024-01-22
Update local Spack package

RMeli:
The local Spack package appears to be behind what is in upstream Spack.


simonpintarelli (approved):
Thanks.

PR #955, opened 2024-01-24
[enchancement] use GSL to solve radial equations

toxa81:
Major refactoring of the radial solver. Custom RK4 integrator is removed in favour of GSL implementation. Bound state energies for Schroedinger and Dirac equations are in good agreement comparing with analytical results. Other tweaks:
* atom utility is refactored
* unused files are deleted
* generation of radial functions is improved; in particular generation of LAPW basis for high orbital quantum numbers is enchanced
* minor refactoring related to Poisson solver


toxa81:
cscs-ci run default

apps/atoms/atom.cpp
```diff
@@ -791,18 +821,18 @@ generate_atom_file(Free_atom& a, double core_cutoff, const std::string& lo_type,
 
     //            fprintf(fout, "        <lo l=\"%i\">\n", l);
```
mtaillefumier: unless there is a reason to keep this, I would suppress the commented code entirely. We can always put it back with git when necessary

src/radial/radial_solver.hpp
```diff
@@ -145,517 +191,397 @@ class Radial_solver
...
+            switch (rel) {
+                case relativity_t::none:
+                case relativity_t::koelling_harmon:
+                case relativity_t::zora: {
+                    double ll_half = p->l * (p->l + 1) / 2.0;
+                    double M       = rel_mass<rel>(p->enu, V);
+                    /* p'(x) = 2 * M(x) * q(x) + p(x) / x + chi_p(x) */
+                    f[0] = 2 * M * y[1] + y[0] / x + chi_p;
+                    /* q'(x) = (V(x) - enu + l * (l + 1) / 2 / M(x) / x^2) * p(x) - q(x) / x + chi_q(x) */
+                    f[1] = (V - p->enu + ll_half / M / x / x) * y[0] - y[1] / x + chi_q;
```
mtaillefumier: I may replace` /M / x / x` with `/ (M * x * x)` or better `/ (M * (x * x))`. Reason we fix the order of the operations.

mtaillefumier: and the division is costly.

src/radial/radial_solver.hpp
```diff
@@ -145,517 +191,397 @@ class Radial_solver
...
+                    break;
+                }
+                case relativity_t::iora: {
+                    double ll_half = p->l * (p->l + 1) / 2.0;
+                    double M       = rel_mass<rel>(p->enu, V);
+                    double M0      = rel_mass<relativity_t::zora>(p->enu, V);
+
+                    f[0] = 2 * M * y[1] + y[0] / x + chi_p;
+                    f[1] = (V - p->enu + ll_half / x / x / M0 - ll_half * sq_alpha_half * p->enu / x / x / M0 / M0) *
```
mtaillefumier: see point above.

mtaillefumier (approved):
It looks alright to me. Since the code use GSL for RK method, it might be useful to give the user the choice about the method. RK4 is a bit borderline, I would use gsl_odeiv2_step_rk8pd as it has better mathematical properties but it might simply not be relevant for these calculations.

toxa81:
> It looks alright to me. Since the code use GSL for RK method, it might be useful to give the user the choice about the method. RK4 is a bit borderline, I would use gsl_odeiv2_step_rk8pd as it has better mathematical properties but it might simply not be relevant for these calculations.

actually, the code does `const gsl_odeiv2_step_type* T = gsl_odeiv2_step_rk8pd;`. I was thinking to add a flexibility of controlling this, but then decided to fix it to `rk8pd`

mtaillefumier:
> > It looks alright to me. Since the code use GSL for RK method, it might be useful to give the user the choice about the method. RK4 is a bit borderline, I would use gsl_odeiv2_step_rk8pd as it has better mathematical properties but it might simply not be relevant for these calculations.
> 
> actually, the code does `const gsl_odeiv2_step_type* T = gsl_odeiv2_step_rk8pd;`. I was thinking to add a flexibility of controlling this, but then decided to fix it to `rk8pd`

Perfect.

mtaillefumier:
I have no objection to merge this PR. Most of my comments are cosmetics

toxa81:
> I have no objection to merge this PR. Most of my comments are cosmetics

I will quickly do the optmisations of the divisions

PR #956, opened 2024-01-24
[enhancement] more robust calculation of radial functions

toxa81:
Main change: try to generate APW radial functions. In case of failure use the previous values. 
 
A few minor fixes in other parts of the code


toxa81:
cscs-ci run default

mtaillefumier (approved):

PR #957, opened 2024-01-27
[test] vector calculus on periodic funcitons

toxa81:
Two expressions for GGA potential both for regular-spaced FFT grid and muffin-tins. This is controlled by switch in the `settings` section. Fixed a few minor issues for complex Smooth_periodic_function.


toxa81:
TODO: laplacian formulation for magnetic case for both regular FFT grid and muffin-tins

simonpintarelli:
I think it's good to have the option for experimenting. 

What about having
```
Smooth_periodic_function to_rg(Smooth_periodic_function&& f)
{
  f.fft_transform(1);
  return f;
}
```
instead of the toggle switch to gradient, laplacian?

simonpintarelli (approved):
lgtm

PR #958, opened 2024-02-01
[API] adding parameter in diagonalize

toxa81:


src/api/sirius.mod
toxa81, line 1: this has to be deleted from PR

src/hamiltonian/diagonalize_pp.hpp
```diff
@@ -315,10 +315,10 @@ diagonalize_pp(Hamiltonian_k<T> const& Hk__, K_point<T>& kp__, double itsol_tol_
 
         std::stringstream s;
         std::ostream* out = (kp__.comm().rank() == 0) ? &std::cout : &s;
-
+        int num_steps = itsol_num_steps__;
```
toxa81: this is redundant; use `itsol_num_steps__` directly

toxa81 (commented):
LGTM

PR #962, opened 2024-02-12
[update] JSON parser

toxa81:
* use latest version of nlohmann/json parser
* tweak pretty print for the arrays: print them in single line


simonpintarelli (approved):
We might also once consider to use it as dependency via spack.

PR #963, opened 2024-02-13
[fix] missing includes

simonpintarelli:
recently added `math_tools.hpp` didn't compile with gcc12.


toxa81:
long term: we need gcc12 ci/cd pipeline

simonpintarelli:
I think in near future we will move the CI to alps/GH, then there won't anything older than gcc12.

PR #964, opened 2024-02-13
[CMake] Intel OneAPI MKL

RMeli:
This PR removes the custom `FindMKL.cmake` module in favor of using the `MKLConfig.cmake` provided by Intel OneAPI MKL (since version `2021.3`).

`intel-mkl` is now deprecated in Spack (see https://github.com/spack/spack/pull/41488), and the last release is `2020.4.30`.

We recently moved to Intel OneAPI MKL in DLA-Future (see https://github.com/eth-cscs/DLA-Future/pull/1073), which broke `sirius+dlaf` since they both use `MKL_FOUND` and other equally named variables, but then use `sirius::mkl` and `MKL::MKL`, respectively. This leads to
```
    179      The link interface of target "DLAF::LAPACK" contains:
     180    
     181        MKL::MKL
     182    
     183      but the target was not found.  Possible reasons include:
```
 
(Thanks @albestro for the help spotting that sharing `MKL_FOUND` was the root of the problem.)


simonpintarelli (approved):
Thank you @RMeli , lgtm.

PR #965, opened 2024-02-13
[feature] save magnetisation in JSON output

toxa81:
Save magnetic moments in the JSON output. All reference files were regenerated with the new version. Total energies were checked to match the previous values.


simonpintarelli (approved):
lgtm

PR #967, opened 2024-02-19
[enhancement] use strong type for G-vector index

toxa81:
* remove index_domain_t and introduce strong type for G-vector index
* introduce iterator for G-vectors
  * also introduce skip_g0() wrapper function to skip G=0 component


simonpintarelli (approved):
lgtm

PR #968, opened 2024-02-22
Python updates

simonpintarelli:


src/context/config.hpp
toxa81, line 1: @simonpintarelli generating script should be updated, config.hpp should not be hand-crafted.

toxa81 (changes requested):

src/context/config.hpp
simonpintarelli: I've reverted it. Seems my previous comment disappeared.

toxa81 (approved):
LGTM

PR #971, opened 2024-03-01
add a separate target for fortran api

simonpintarelli:
q-e-sirius currently fails in cmake configure when `intel-oneapi-mkl` is used, because `MKL_INTERFACE`, `MKL_THREADING` aren't propagated to q-e-sirius spack recipe.

The Fortran API actually doesn't require to find SIRIUS dependencies in cmake, as they are private dependencies (assuming shared library is used).

This PR changes cmake config files as follows:
- `sirius_cxxConfig.cmake` the C++ library, using the hand-written config file as it was before
- `siriusConfig.cmake` is now intended only for the Fortran API. 

The target `sirius::sirius` is the F90 API, using `sirius::sirius_cxx` as private dependency.

The change should be backwards compatible, since the C++ API isn't used in any project.


src/CMakeLists.txt
```diff
@@ -86,24 +86,23 @@ set(_SOURCES
...
-# create library with .cpp, .cu and .f90 sources
-add_library(sirius "${_SOURCES};${_CUSOURCES};${_FSOURCES}")
-
-target_compile_features(sirius PUBLIC cxx_std_14)
-target_compile_features(sirius PUBLIC c_std_99)
-target_compile_features(sirius PUBLIC cuda_std_14)
-set_target_properties(sirius PROPERTIES POSITION_INDEPENDENT_CODE ON)
+  # create library with .cpp, .cu and .f90 sources
+add_library(sirius_cxx "${_SOURCES};${_CUSOURCES};")
+target_compile_features(sirius_cxx PUBLIC cxx_std_14)
```
RMeli:
```suggestion
target_compile_features(sirius_cxx PUBLIC cxx_std_17)
```
Isn't SIRIUS using C++17 now?

src/CMakeLists.txt
```diff
@@ -86,24 +86,23 @@ set(_SOURCES
...
+target_compile_features(sirius_cxx PUBLIC c_std_99)
+target_compile_features(sirius_cxx PUBLIC cuda_std_14)
```
RMeli:
```suggestion
target_compile_features(sirius_cxx PUBLIC cuda_std_17)
```

src/api/sirius.f90
RMeli, line 1: Is this file part of this PR?

src/api/sirius_api.cpp
RMeli, line 1: Is this file part of this PR?

src/CMakeLists.txt
simonpintarelli: done

src/api/sirius.f90
simonpintarelli: It isn't, fixed by rebase.

RMeli (approved):

PR #976, opened 2024-03-16
python updates

simonpintarelli:
- fix / cleanup of nlcg python
- add new pytest


src/unit_cell/unit_cell.hpp
```diff
@@ -329,6 +331,17 @@ class Unit_cell
         return *atom_types_[id__];
     }
 
+    inline bool
+    has_augmented_atom() const
```
toxa81: Unit_cell::augment() provides same functionality

toxa81:
@simonpintarelli please have a look at the comment about `has_augmented_atom()`. Otherwise it's good to go

simonpintarelli:
@toxa81 which comment?

src/unit_cell/unit_cell.hpp
toxa81: @simonpintarelli here

toxa81 (changes requested):

src/unit_cell/unit_cell.hpp
simonpintarelli: Thanks. I've removed my duplicate.

toxa81 (approved):
LGTM

PR #979, opened 2024-03-31
[fix] compute free energy

toxa81:


toxa81:
@simonpintarelli [not urgent] Hi Simon, can you have a look when you have time?

PR #980, opened 2024-04-04
Enable direct reading of UPF version >= 2 PP files

abussy:
This PR enables the direct reading of standard pseudopotential UPF files with version >= 2, without the need for the `upf_to_json` tool. The `pugixml` library is used for this purpose.

All functions reading information from JSON PP files, namely `Atom_type::read_pseudo_uspp` and `Atom_type::read_pseudo_paw` have been overloaded with function reading from xml files (stepwise identical). The relevant function is called depending on the file termination in the SIRIUS input file.

This PR is fully backward compatible. At this stage, the dependency on the `pugixml` library was made a hard requirement. This would imply a change in the spack recipe for the next release.

**Edit:** at this stage, it is an open question whether `pugixml` should be a hard requirement or an option.


abussy:
The `pugixml` dependency was made optional with the addition of a spack variant `+pugixml`. Whether `pugixml` is enabled or not, json potential files can be parsed. This PR is fully backward compatible.

Note: for use of `pgixmp` with CP2K, the CP2K spack recipe (and probably the CP2K toolchain) will need to be adapted. Moreover, the CP2K code will need a slight refactorization, as it currently assumes that SIRIUS only reads JSON files, and it falls back to its own (incomplete) UPF v2 parser otherwise: https://github.com/cp2k/cp2k/blob/2ed94b8b96b2a4590f5340063d84c5d639a6fefe/src/atom_upf.F#L120-L127.

simonpintarelli:
Just as a comment: SSSP currently still contains many UPFv1 pseudos (GRBV), but in the future it will be all UPFv2 🎉 .
The remaining GRBV UPFv1 will be converted to v2 using https://github.com/azadoks/PseudoPotentialIO.jl.

simonpintarelli:
> Edit: at this stage, it is an open question whether pugixml should be a hard requirement or an option.

Imho as long as QE parses the UPFs, `pugixml` should be a variant. The question is if we want to rely on QE to parse UPFs in q-e-sirius or not. At least until a future SSSP is released only containing UPFv2, we should keep it as an optional dependency.

simonpintarelli:
Adding a test under `verification` would be useful too. E.g. first generate an `output_ref.json` using the native json format and then run the same input using UPF files directly.

src/unit_cell/atom_type.cpp
```diff
@@ -737,6 +733,338 @@ Atom_type::read_input(std::string const& str__)
...
+    pugi::xml_node nl_node = upf.child("PP_NONLOCAL");
+    if (!nl_node.child("PP_AUGMENTATION").empty()) {
+        for (int i = 0; i < nbf; i++) {
+            std::string istr     = "PP_BETA." + std::to_string(i + 1);
+            pugi::xml_node inode = nl_node.child(istr.data());
+            int li               = inode.attribute("angular_momentum").as_int();
+
+            for (int j = i; j < nbf; j++) {
+                std::string jstr     = "PP_BETA." + std::to_string(j + 1);
+                pugi::xml_node jnode = nl_node.child(jstr.data());
```
unkcpz, line 882: (@simonpintarelli redirect me to this PR. I am maintaining SSSP and working on update SSSP with some improvement to v2) 
I think in order to compatible with GBRV pseudos, `nqf!=0` should also be supported here. And GBRV pseudos has q_ij rather than q_ijl, https://github.com/QEF/q-e/blob/57a97fe3ac56c6862ee083bf10d27505835c092a/upflib/read_upf_v1.f90#L65-L67

abussy: As far as I can tell, GBRV pseudos are stored as UPF version 1 files. We are not aiming to write such a parser. If you are planing to convert to GBRV pseudos to v2, we could cover them as well. But then, I assume they would follow the UPF v2 standard?

simonpintarelli:
cscs-ci run default

src/unit_cell/atom_type.cpp
unkcpz: > But then, I assume they would follow the UPF v2 standard?

Yes, but the UPF v2 standard is for the format, the non-zero `nqf` and `q_ij` are still the valid fields according to the UPF v2 schema and according to the notes provided in https://github.com/QEF/qeschemas/pull/11#issuecomment-1238139482

toxa81: As far as I remember V2 UPF pseudos introduced Q_ij^\ell(r) radial functions as a hard standard. Long term it is better to conver V1 GBRV to V2.

unkcpz: https://gist.github.com/unkcpz/c3287268c84ba46b3366edb5537d65c7

Here is a Pd pseudo I convert from UPF.v1 GBRV to UPF.v2. In the QE, it will then be read by [read_upf_new.f90](https://github.com/QEF/q-e/blob/develop/upflib/read_upf_new.f90) instead of fallback read by [read_upf_v1.f90](https://github.com/QEF/q-e/blob/develop/upflib/read_upf_v1.f90). 
It is true from Giannozzi's note (https://github.com/QEF/qeschemas/pull/11#issuecomment-1238139482) that `qfcoef` and `rinner` are to be considered obsolescent. The plan for the SSSP v2 is to provide GBRV with converting it to UPFv2 format (let me know if this is useful, I need more test on such UPFv2 file I converted), I assume in SIRIUS you still can use it? But as @toxa81 pointed, maybe in the long term should push to have use uniformed expression.

abussy: I personally think that having the whole SSSP library in UPF v2 format would be very useful, especially for people outside of QE. I also think that having a single uniform standard is important. For example, in this case, we would have to implement a special case for `qfcoef` and `rinner`, which in the end, is not very different from having 2 distinct parsers.

As far as I understand, it is possible to go from UPF v1 `PP_QIJ` to UPF v2 `PP_QIJL` seamlessly. For example, the `upf_to_json` tool of SIRIUS parses both [UPF v1](https://github.com/electronic-structure/SIRIUS/blob/bd9bb4581a9ee6939f89001b1336e29859ced84c/apps/upf/upf1_to_json.py#L247-L354) and [UPF v2](https://github.com/electronic-structure/SIRIUS/blob/bd9bb4581a9ee6939f89001b1336e29859ced84c/apps/upf/upf2_to_json.py#L119-L152) to the same JSON structure, which happens to be very close to that of UPF v2 (`PP_QIJL`).

unkcpz: > For example, the upf_to_json tool of SIRIUS parses both [UPF v1](https://github.com/electronic-structure/SIRIUS/blob/bd9bb4581a9ee6939f89001b1336e29859ced84c/apps/upf/upf1_to_json.py#L247-L354) and [UPF v2](https://github.com/electronic-structure/SIRIUS/blob/bd9bb4581a9ee6939f89001b1336e29859ced84c/apps/upf/upf2_to_json.py#L119-L152) to the same JSON structure 

That is cool, I didn't know it actually convert not only the format but the expression as well. I'll talk to @simonpintarelli to see how it can be realize and I can probably add it to my v1 -> v2 converter.

simonpintarelli:
cscs-ci run default

simonpintarelli:
cuda container is failing on non-existent +pugixml variant:
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/7415069138667150/4626796664769983/-/jobs/6591482684#L6282

🤔 the cscs-ci container takes the `sirius/package.py` from upstream spack, while in github-ci we use the `package.py` from the repository.

simonpintarelli:
cscs-ci run default

(why does it keep disappearing?)

simonpintarelli:
Ping @finkandreas we are running into a timeout when building the cuda base container:
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/7415069138667150/4626796664769983/-/jobs/6594225309#L6216

Should we skip it until rosa is back?

finkandreas:
or increase timeout to e.g. 3h, 4h, 5h. No idea how long it takes, but the runner allows long runtimes. 
Should be this line: https://github.com/electronic-structure/SIRIUS/blob/develop/ci/cscs-daint.yml#L12

simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

abussy (commented):
It's getting out of hand :)

src/unit_cell/atom_type.cpp
unkcpz: Just quick update from my side. The converter are ready to convert to PP_QIJL and did exactly the same as what happened in QE internally. So it will be safe with this implementation with the new SSSP 2.0 in the future.

PR #987, opened 2024-04-18
Fix test32

abussy:
Regenerated a reference output file for `test32`. Error originated from PR #980 and #979 being merged in close succession.


simonpintarelli:
cscs-ci run default

PR #996, opened 2024-05-22
[fix] performace of symmetrize_mt_function()

toxa81:
Optimise symmetrization of muffin-tin functions. The following steps are taken:

- Rlm rotation matrices are cached in simulation contexts. They were somehow expensive to generate each time
- symmstrizaion is now performed for each atom class independently on an optimal 2D MPI grid. First dimension is used to parallelize atoms of the given symmetry class, second dimension is used to parallelize radial grid points. As such, all MPI rank are involved in symmetrization
- minor fix in initializing muffin-tin magnetization - it was not synchronised between all MPI ranks

### DIFF
--- a/src/context/simulation_context.cpp
+++ b/src/context/simulation_context.cpp
@@ -815,7 +815,7 @@ Simulation_context::update()
     unit_cell().update();
 
     /* cache rotation symmetry matrices */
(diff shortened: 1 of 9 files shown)


toxa81:
cscs-ci run default

simonpintarelli (approved):
lgtm

PR #998, opened 2024-05-27
Remove hard-coded printing of pika bindings

RMeli:
Printing pika bindings was hard-coded. It is too verbose with many ranks (and many cores) to force it to the user. This PR removes the printing. It can be enabled by the user by exporting the `PIKA_PRINT_BIND` variable.


RMeli:
cscs-ci run

Issue #196, opened 2018-04-09
get rid of 10a.u. hack in sirius


toxa81:
This is done to match QE results. Better to check for proper analytical tail of rho_core(r) and vloc(r)

10 a.u. is hardcoded and used as `rcut` in  Modules/read_pseudo.f90

toxa81:
turned into input parameter

PR #1002, opened 2024-06-12
Added PP radial grid cutoff as input paramter

abussy:
Currently, in SIRIUS, there is a hard coded cutoff of 10 a.u. on the PP radial grids for integration. This is a protection against badly generated PPs that do not have a properly vanishing tail. The exact value is inherited from Q.E. (see PR #196 ).

Other codes may use different values for this. There are also codes, such as DFTK, which allow passing this value as an input parameter.

This PR introduces the cutoff as an input parameter. The default value is 10 a.u., which insures back compatibility. The `test04` was modified to include a non-default grid cutoff.


toxa81:
cscs-ci run default

toxa81:
cscs-ci run

PR #1006, opened 2024-06-17
[fix] Fixed bug in direct XML parsing of UPF v2 PPs

abussy:
Fixed bug when parsing UPF files without NLCC. When there is no core correction, the default `json` behavior is to allocate an array for it filled with zeros. So far, the `pugixml` parser gave a size zero vector, leading to a crash. This PR fixes it.


toxa81:
cscs-ci run default

toxa81:
cscs-ci run

PR #1007, opened 2024-06-24
[enhancement] cleanup of tests and unit tests

toxa81:
* clean-up of tests and unit tests
* old and obsolete tests are removed
* use initializer_list in cmd_args
* uniform way to call tests
* two verification tests were updated to have more accurate ground state


toxa81:
cscs-ci run

mtaillefumier (approved):
looks good

PR #1008, opened 2024-07-07
Add timers for Hubbard symmetrization

simonpintarelli:
Add `sirus::symmetrize_occupation_matrix`. 

@gsavva encountered a SCF+U calculation which spent the bulk of the time in `symmetrize_occupation_matrix` in `Density::generate`.

### DIFF
--- a/src/hubbard/hubbard_matrix.hpp
+++ b/src/hubbard/hubbard_matrix.hpp
@@ -240,6 +240,7 @@ class Hubbard_matrix
     auto
     find_orbital_index(const int ia__, const int n__, const int l__) const
     {
+        PROFILE("Hubbard_matrix::find_orbital_index");
         for (int at_lvl = 0; at_lvl < static_cast<int>(atomic_orbitals_.size()); at_lvl++) {
             int lo_ind = atomic_orbitals_[at_lvl].second;
 
--- a/src/hubbard/hubbard_matrix.hpp
+++ b/src/hubbard/hubbard_matrix.hpp
@@ -240,7 +240,6 @@ class Hubbard_matrix
     auto
     find_orbital_index(const int ia__, const int n__, const int l__) const
     {
-        PROFILE("Hubbard_matrix::find_orbital_index");
         for (int at_lvl = 0; at_lvl < static_cast<int>(atomic_orbitals_.size()); at_lvl++) {
             int lo_ind = atomic_orbitals_[at_lvl].second;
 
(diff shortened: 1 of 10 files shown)


toxa81:
I'm almost sure it's the same issue that we solved in symmetrize_mt_function(): generation of `sht::rotation_matrix<double>(4, eang, pr);` is expensive and needs to be cached. `symmetrize_mt_function()` already contains the fix

simonpintarelli:
There is some room for improvment here ^^. I've cherry picked the changes and compiled on todi, At least it seems to fix the bulk of the performance issue.

src/symmetry/symmetrize_occupation_matrix.hpp
```diff
@@ -46,6 +48,19 @@ symmetrize_occupation_matrix(Occupation_matrix& om__)
         }
     }
 
+    std::vector<std::vector<mdarray<double, 2>>> rotms(sym.size());
```
toxa81: but why not passing it as argument? it is already computed in sim.ctx.

simonpintarelli: Aren't the ones in sim ctx different? https://github.com/electronic-structure/SIRIUS/blob/sirius-md/src/context/simulation_context.cpp#L834

It does the following at the end:
https://github.com/electronic-structure/SIRIUS/blob/sirius-md/src/core/sht/sht.cpp#L98

In `symmetrize_occupation_matrix` for each symmetry 
https://github.com/electronic-structure/SIRIUS/blob/sirius-md/src/core/sht/sht.cpp#L115
was called, which returns a vector of rotations matrices.

I didn't look at the for loops below, perhaps one can rearrange things and use the ctx.rotm().

toxa81: I checked the code and I found a TODO that perfectly fits this refactoring. I'll add my changes to this PR

src/hubbard/hubbard_matrix.hpp
```diff
@@ -240,6 +240,7 @@ class Hubbard_matrix
     auto
     find_orbital_index(const int ia__, const int n__, const int l__) const
     {
+        PROFILE("Hubbard_matrix::find_orbital_index");
```
simonpintarelli: this line is malicious, the profiler is not thread safe ...

toxa81:
cscs-ci run

toxa81:
@simonpintarelli @gsavva is it good to be merged?

PR #1016, opened 2024-08-08
[WIP] API changes for Julia bindings

abussy:
This PR brings some changes to the API, in order to facilitate the creation of Julia bindings for SIRIUS in a future `SIRIUS.jl` package.

Apart from the additions of new functions to `sirius_api.cpp` (necessary for the `DFTK` integration), the main addition consists of a header file for the API: `sirius_c_headers.h`. It turns out that having a header file drastically simplifies the process of creating Julia bindings for a C/C++ library.

As things stand, this header file is generated by a script (`generate_c_headers.py`), in the same spirit as the Fortran API. As a result, **it is not used for compilation**, but only as source material for the automatic generation of Julia bindings.

The above point is why this PR is marked as WIP. I believe we might want to have a discussion as to whether we follow this direction, or if we decide to have a proper header file that we also use for compilation.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -157,6 +157,8 @@ if(SIRIUS_CREATE_FORTRAN_BINDINGS)
   set_target_properties(sirius PROPERTIES Fortran_MODULE_DIRECTORY mod_files)
   install(FILES "${PROJECT_BINARY_DIR}/src/mod_files/sirius.mod"
           DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/sirius")
+  install(FILES "${PROJECT_SOURCE_DIR}/src/api/sirius_c_headers.h"
+          DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/sirius/src/api")
   install(TARGETS sirius
     EXPORT sirius_targets
     ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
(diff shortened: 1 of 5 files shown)


toxa81 (approved):
To me this looks fine. There are no breaking changes, and .h file serves its purpose for Julia bindings. Later we can get back to a discussion how to improve it.

abussy:
Any opinion on this @simonpintarelli, @mtaillefumier ?

mtaillefumier:
No major objections especially if the header file generated automatically.

simonpintarelli:
The docstrings are now duplicated in the header and in `sirius_api.cpp`. This can be cleaned up, docstrings should be in the header and then the rest adapted accordingly to generate sirius.f90 from the header. Fine for me to do it in a future PR.

toxa81:
> The docstrings are now duplicated in the header and in `sirius_api.cpp`. This can be cleaned up, docstrings should be in the header and then the rest adapted accordingly to generate sirius.f90 from the header. Fine for me to do it in a future PR.

Good catch @simonpintarelli ! However, I would still use sirius_api.cpp as a source of truth for the documentation and a bare header file only for Julia bindings. @abussy will it be possible to cleanup .h file from the doc. comments?

abussy:
My script to generate Julia bindings makes use of the docstrings. I am also not sure I could get the data from `sirius_api.cpp`, as it is not shipped in the `include` folder of the library (and putting it there would be akward).

PR #1022, opened 2024-09-19
fix rocm include for rocm v6

simonpintarelli:
rocblas has changed the interface for `trmm` in v6.0.0. 

**TODO** either make the change backward compatible, or upgrade rocm to 6.0.0 in the base container too.


toxa81:
IMO better to move forward and update base container

PR #1025, opened 2024-11-02
Make sirius compatible with libxc 7

mtaillefumier:
- Libxc 7 introduces a breaking change in their header file. The xc functionals are now defined in a separate header file `xc_funcs.h` that is not included by default in `xc.h`. 
- Apply formatting

### DIFF
--- a/python_module/py_sirius.cpp
+++ b/python_module/py_sirius.cpp
@@ -468,9 +468,8 @@ PYBIND11_MODULE(py_sirius, m)
 
     py::class_<fft::Grid>(m, "FFT3D_grid")
             .def_property_readonly("num_points", py::overload_cast<>(&fft::Grid::num_points, py::const_))
-            .def_property_readonly("shape", [](const fft::Grid& obj) -> std::array<int, 3> {
-                return {obj[0], obj[1], obj[2]};
-            });
+            .def_property_readonly("shape",
+                                   [](const fft::Grid& obj) -> std::array<int, 3> { return {obj[0], obj[1], obj[2]}; });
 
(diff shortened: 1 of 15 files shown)


mtaillefumier:
apply-format.x does not seem to format the code correctly on my machine....

toxa81:
> apply-format.x does not seem to format the code correctly on my machine....

Probably because behaviour of clang-format is slightly different from version to version. You need to use llvm@17.0.4. Also, format should not be applied to files that are untouched in the current PR. 

BTW, is the change backwards compatible with libxc-6.x?

mtaillefumier:
It is compatible with libxc 6. The missing header file is surrounded by a `#if`  statement. I also did not change the spack recipe so if we enforce libxc 6 (which I think we should keep for the time being) then we are safe.

simonpintarelli:
The change regarding libxc looks good to me. We should revert the formatting commit and fix `.clang-format` afterwards. I'm getting the same diff, some changes look fine, others not, for example the indent in the python bindings.

mtaillefumier:
> The change regarding libxc looks good to me. We should revert the formatting commit and fix `.clang-format` afterwards. I'm getting the same diff, some changes look fine, others not, for example the indent in the python bindings.

Oh, I only use the apply-format script. Something is definitely wrong.  I will revert the patch.

simonpintarelli:
I'm trying to figure out what has changed in clang-format, one option which apparently changed defaults is `AlignTrailingComments`.

simonpintarelli (approved):
lgtm

PR #1026, opened 2024-11-04
Update to clang format 18

simonpintarelli:
- Update `clang-format` in github actions from 17 to 18
- reformat the code (didn't manage to tweak the option such that there is no diff between the two versions)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -26,6 +26,7 @@ UseTab: Never
 AllowShortFunctionsOnASingleLine: 'None'
 AllowShortIfStatementsOnASingleLine : 'false'
 AllowShortLoopsOnASingleLine: 'false'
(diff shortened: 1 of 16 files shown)


toxa81:
It's OK to re-apply the format, but it should be done in a standalone PR.

simonpintarelli:
I'm trying to figure out the options to make the least changes in a reformat.

simonpintarelli:
@mtaillefumier could you please check if this is consistent with clang-format of llvm 17?

mtaillefumier:
I will do that right away

mtaillefumier (approved):
I applied the formating script locally and no change to the files. This seems to fix the issue I had last week. Thanks.

Approved

toxa81 (approved):
OK to me

PR #1031, opened 2024-11-25
[API] pass data back to QE

gsavva:
Implementation of more API functions to pass data from SIRIUS to QE:
- density matrix
- local and non-local occupation matrix

Linked to [PR#69](https://github.com/electronic-structure/q-e-sirius/pull/69) on the `q-e-sirius` side

### DIFF
--- a/src/api/sirius.f90
+++ b/src/api/sirius.f90
@@ -6303,32 +6303,37 @@ deallocate(file_name_c_type)
 end subroutine sirius_load_state
 
 !
-!> @brief Set density matrix.
+!> @brief Access (get or set) density matrix.
 !> @param [in] gs_handler Ground-state handler.
+!> @param [in] access_type Access type ("get" or "set").
 !> @param [in] ia Index of atom.
-!> @param [in] dm Input density matrix.
+!> @param [inout] dm Input density matrix.
 !> @param [in] ld Leading dimension of the density matrix.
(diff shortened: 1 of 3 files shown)


toxa81:
I have one proposal to consider: how about making a common "access" function that get/sets matrices in a single place? For example:
```
if (access == "get") {
  qe_matrix = sirius_matrix;
} else {
sirius_matrix=qe_matrix;
}
```
 and the structure of the `for` loops over l,m indices will remain the same

gsavva:
@toxa81 thanks for the suggestion on implementing a single "access" function.
I've completed all the necessary changes, both here and in `q-e-sirius` side ([PR#69](https://github.com/electronic-structure/q-e-sirius/pull/69))
It remains to be tested thoroughly...

gsavva:
FYI, the part for the non-local occupation matrix has a bug.
A local test, scf+U+V, runs and converges as expected but [this error](https://github.com/electronic-structure/SIRIUS/blob/feature/more_API_functions/src/api/sirius_api.cpp#L6731) is given out.
I'm investigating it...

gsavva:
After some fixes on the q-e-sirius side, this should be good to merge.
I've done a few tests locally and the restart works in the case of scf+U+V.
@toxa81 @mtaillefumier please review this PR

toxa81 (approved):
looks fine

PR #1032, opened 2024-11-27
[refactoring] minor changes to the source code

toxa81:
* use common naming convention for Gaunt coefficients
* fix return type of <R|Y|R> Gaunt coefficients

ping @mtaillefumier for review

### DIFF
--- a/src/core/sht/sht.hpp
+++ b/src/core/sht/sht.hpp
@@ -266,24 +266,6 @@ class SHT // TODO: better name
         }
     }
 
-    /// Convert from Ylm to Rlm representation.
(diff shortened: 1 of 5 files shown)


toxa81:
cscs-ci run default

toxa81:
@mtaillefumier <R|Y|R> gaunt coefficients were treated as real. Made them complex. This is used in one place in SO code, can you please check?

mtaillefumier (approved):
We use the Clebsh Gordan coefficients in the SO case so I do not think this refactoring should affect SO calculations at all.

PR #1033, opened 2024-11-28
[fix] Gamma-point collinear case

toxa81:
There was too strict check of the spin component. Fixed now.


toxa81:
Strage.. I can run all tests on my Linux workstation

toxa81:
cscs-ci run default

simonpintarelli (approved):
lgtm

PR #1034, opened 2024-11-28
touch file

toxa81:


toxa81:
@simonpintarelli @mtaillefumier I don't get this. Last merged PR to develop is greenlighted. All was good. I make a single character change, and this PR fails.

toxa81:
cscs-ci run default

simonpintarelli:
I'm getting the same diff,  clang-format removes the additional newline in `src/_header_file_template.hpp`

toxa81:
> I'm getting the same diff, clang-format removes the additional newline in `src/_header_file_template.hpp`

yes, it's just to introduce a difference so CI can be triggered.

simonpintarelli:
Ah ok, I was only looking at the formatting change. 

I'll try to run it locally. The unit tests are failing with SIGPIPE.

simonpintarelli:
ok, after undoing and reapplying the whitespace commit :), it looks like a glitch with sarus which caused the PIPE error.

PR #1037, opened 2024-12-11
[refactor] valence D-matrix in Potential class

toxa81:
Logically, D-operator matrix must be part of the Potential class as it represent non-local part of pseudopotential. It is already generated in Potential, but was stored in Atom for no good reason.


simonpintarelli (approved):
lgtm

PR #1039, opened 2024-12-18
Add gsl include directory at compilation

mtaillefumier:
cmake searches for GSL but the target is used nowhere leading to a compilation error. GSL::gsl should be added to sirius_cxx to fix the potential issue


simonpintarelli (approved):
lgtm
I wonder why it was linking gsl correctly in spack?

I double checked that it doesn't link gsl cblas. In the current cmake this is a separate target, iirc this wasn't always the case.

PR #1041, opened 2024-12-19
[refactor] core density is computed in Density class

toxa81:
The idea is that the unit cell, atoms and atom classes should be immutable once the simulation context is created. Right now this is not the case as atoms and atom symmetry classes are used to generate radial functions, radial integrals and core density. This is the first step: move generation of core density to Density class itself.


mtaillefumier (approved):
this PR seems fine.

PR #1044, opened 2024-12-29
[fix] non-collinear magnetisation

toxa81:
- fix for fp-lapw case - non-diagonal block of density matrix is computed from conj(psi_dn) * psi_up
- experimental fix for PAW - do not use initial density matrix


toxa81:
cscs-ci run default

mtaillefumier (approved):

PR #1048, opened 2025-01-15
[fix] u-matrix

toxa81:


toxa81:
cscs-ci run default

toxa81:
@mtaillefumier just a reminded that a fix on QE side is also equired. You can't pass J like this `J=Hubbard_J(1,iat) / 2.0,`
```
             CALL sirius_set_atom_type_hubbard(sctx, TRIM(atom_type(iat)%label), &
                   & l=Hubbard_l(iat), n=Hubbard_n(iat), occ=Hubbard_occ(iat, 1), &
                   & U=Hubbard_U(iat) / 2.0, J=Hubbard_J(1,iat) / 2.0, &
                   & alpha=Hubbard_alpha(iat) / 2.0, beta=Hubbard_beta(iat) / 2.0, &
                   & J0=Hubbard_J0(iat) / 2.0)

```
I believe there is a mismatch between how you describe J and treat it

```
    J:
      type: double
      attr: in, required
      doc: Exchange J parameter for the full interaction treatment.

```

mtaillefumier:
I changed the internal data structure for the hubbard coefficients to match QE better.

toxa81:
@mtaillefumier  J__ now is double[3] array, correct?

mtaillefumier:
> @mtaillefumier J__ now is double[3] array, correct?

QE J__ are double array of three elements. SIRIUS does the same now as many of the parameters have a dedicated variable (U, J, alpha, beta). The J__ array is for B, E2, E3. I used the vector class for this to simplify things a bit.

toxa81:
> > @mtaillefumier J__ now is double[3] array, correct?
> 
> QE J__ are double array of three elements. SIRIUS does the same now as many of the parameters have a dedicated variable (U, J, alpha, beta). The J__ array is for B, E2, E3. I used the vector class for this to simplify things a bit.

I changed to std::array<double,3> and also changed the API to reflect that J is double[3] array.

toxa81:
cscs-ci run default

toxa81:
```
+ spack config --scope=user add config:build_stage:/dev/shm/spack-stage
==> Error: The "config" section of the configuration needs to be written to disk, but is currently using a deprecated format. Please update it using:
	spack config [--scope=<scope>] update config
Note that previous versions of Spack will not be able to use the updated configuration.
```
@simonpintarelli  do you know, why it is breaking?

toxa81:
cscs-ci run default

PR #1051, opened 2025-01-21
Update xc_functional.hpp

mtaillefumier:
add libxc 7 support


mtaillefumier:
already present. spack was not updated.

PR #1053, opened 2025-01-22
UENV CI

simonpintarelli:
Since the container CE fails on a second MPI_Init (ctest calls all test in the same srun) I suggest to use uenv for the CI. 

- cuda tests are run on alps daint
- cpu tests are run on eiger

stages:
- base image: build sirius@develop with same specs as needed later
- build: use baseimage as spack upstream, spack build stage in /dev/shm -> store spack build stage for sirius in artifact
- test: unpack artifact again to /dev/shm and run ctest there

The build stage previously built a container based on the baseimage and then pushed it to jfrog (incl the build directory). Now we store the build directory as artifact and unpack it before running the tests. Therefore I think it's around the same cost.

**Note** I disabled auto trigger for cscs ci, it must be run manually now. I think otherwise it would waste a lot of NH.


simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

toxa81 (approved):
LGTM

PR #1054, opened 2025-01-28
[fix] set explicit row-major order

toxa81:
Explicitly set row-major MPI rank distribution in the blacs grid.


toxa81:
cscs-ci run default

toxa81:
This is weird. I can run test09 using `srun -N1 -n4 -c20 ...` on Daint. I'm changing ctest to output errors to stdout directly.

src/potential/generate_d_mtrx.cpp
```diff
@@ -65,9 +65,7 @@ Potential::generate_d_mtrx()
         if (!atom_type.augment() || atom_type.num_atoms() == 0) {
             for (int iv = 0; iv < ctx_.num_mag_dims() + 1; iv++) {
                 for (int i = 0; i < atom_type.num_atoms(); i++) {
-                    int ia     = atom_type.atom_id(i);
-                    auto& atom = unit_cell_.atom(ia);
-
+                    int ia = atom_type.atom_id(i);
```
mtaillefumier: very minor comment. We should put ia const for our own sake.

apps/tests/test_wf_inner.cpp
```diff
@@ -15,6 +15,7 @@ int
 test_wf_inner_impl(std::vector<int> mpi_grid_dims__, double cutoff__, int num_bands__, int bs__, memory_t mem__)
 {
     spla::Context spla_ctx(is_host_memory(mem__) ? SPLA_PU_HOST : SPLA_PU_GPU);
+    spla_ctx.set_tile_size_gpu(2096);
```
mtaillefumier, line 19: It is maybe good to have a short comment about this line because we will forget in one month why this value is chosen.

mtaillefumier (approved):
Very minor comments. We can merge this as soon as the ci/cd gives us the green light

src/potential/generate_d_mtrx.cpp
toxa81: resolved!

mtaillefumier (approved):

apps/tests/test_wf_inner.cpp
toxa81: added

PR #1058, opened 2025-02-19
[refactor] improve core generation in LAPW

toxa81:
* free atom density is stored and used to reconstruct the long-range tail of the spherical potential
* more conservative approach to compute linearization energies

Other by-product tweaks:
* less verbose output for the case of many k-points
* save core leakage in the output dictionary
* small fix in lattice scaling in the case of EoS runs


toxa81:
cscs-ci run default

mtaillefumier (approved):
This PR can be merged.

toxa81:
cscs-ci run default

PR #1060, opened 2025-03-17
[fix] more robustness in Enu search

toxa81:
The idea is to search for the top of the band energy. This is always possible by checking for (n-l-1+1) number of nodes in the radial function. Then, to remove degeneracies of radial functions, Enu is adjusted further.


toxa81:
cscs-ci run default

toxa81:
cscs-ci run default

PR #1062, opened 2025-03-18
Fix compilation with clang 19

simonpintarelli:
- fix compilation with clang19
- use `std::make_unique` instead of `std::unique_ptr<..>(new..)`


simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

toxa81 (approved):
lgtm

PR #1063, opened 2025-03-24
[fi] robust Enu

toxa81:
Fix the way linearization energies are searched. First, a top of the band is searched by checking n-l number of nodes. Then the bottom of band is searched. In addition, linearization energy is tweaked during calculation of radial functions.


toxa81:
cscs-ci run default

toxa81:
cscs-ci run default

mtaillefumier (approved):
good to be merged

PR #1065, opened 2025-04-01
[feature] fixed magnetisation

toxa81:
Use 2 Fermi energy levels in collinear case to compute fixed magnetization.


toxa81:
cscs-ci run default

toxa81:
cscs-ci run default

mtaillefumier (approved):
We can merge this PR. no objections on my side with these changes. 
- cp2k will automatically know about the new option so we have nothing to do on that side. The cp2k interface will need some updates. I will create a dedicated branch and open the PR after publishing the new version.

mtaillefumier:
i was interrupted in the middle of the process. Thanks for going forward with it.

mtaillefumier:
we also need to update spack.

PR #1066, opened 2025-04-04
[fix] number of mpi grids created

toxa81:
`mpi grid <-> atom class index` is a redundant mapping. The minimal and sufficient mapping is between mpi grid and number of atoms in the symmetry class.

### DIFF
--- a/src/context/simulation_context.cpp
+++ b/src/context/simulation_context.cpp
@@ -848,7 +848,7 @@ Simulation_context::update()
         namax = std::max(namax, unit_cell().atom_symmetry_class(ic).num_atoms());
     }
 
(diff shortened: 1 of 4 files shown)


toxa81:
cscs-ci run default

toxa81:
@RMeli is it good to be merged?

RMeli:
I was unable to test it because of the GB runs. But I think we can merge it, and I'll test it next week.

toxa81:
This should work, as I tested the initialization on the LiCoO2 ~1500 atom example. This is actually when I discovered the problem.

RMeli (approved):

RMeli:
Indeed, it does work well. Thanks @toxa81!

PR #1077, opened 2025-05-13
[fix] setter/getter functions should be overloaded, not .def_property_readonly

toxa81:


toxa81:
cscs-ci run default

simonpintarelli:
cscs-ci run default

simonpintarelli:
I've defined it now as getter and setter. `.def` creates a function which needs to be called with an argument.

simonpintarelli:
cscs-ci run default

simonpintarelli (approved):
lgtm

toxa81:
cscs-ci run default

PR #1081, opened 2025-07-09
Fixes for vdw and fmt libraries

mtaillefumier:
This PR includes major fixes:

- vdw MPI issue is finally fixed. MPI_all_reduce was missing when calculating the energy contribution of the vdw functional
- Added a message indicating that C_LDA correction is needed when libvdwxc functional are used. 
- XC functionals can be weighted. It is set to one by default
- Compared SIRIUS+libvdwxc to native qe:
    -  Difference in total energy < 1 mHa 
    -  Forces are also correct
    - Stress tensor difference between QE and SIRIUS below 1kbar
- fmt header also fixed


mtaillefumier:
cscs-ci run default

simonpintarelli:
Thanks @mtaillefumier!

simonpintarelli (approved):
lgtm

simonpintarelli:
cscs-ci run default

simonpintarelli:
daint gpu band/k-point parallel tests are failing in test35
```
204: total energy is different: -158.617337 computed vs. -158.616928 reference
```

https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/7415069138667150/4626796664769983/-/jobs/10627183108#L2653

On eiger it runs into walltime.

Ref: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/7415069138667150/4626796664769983/-/pipelines/1917097357

simonpintarelli:
cscs-ci run default

mtaillefumier:
> test35 gave
> 
> ```
> 204: total energy is different: -158.617337 computed vs. -158.616928 reference
> ```

I also saw it. I still can not figure out why including a header file would change the result though.

simonpintarelli:
> > test35 gave
> > ```
> > 204: total energy is different: -158.617337 computed vs. -158.616928 reference
> > ```
> 
> I also saw it. I still can not figure out why including a header file would change the result though.

I don't think the header file did it ..

mtaillefumier:
I also have a segfault on mi300 which I do not fully understand yet either.

mtaillefumier:
cscs-ci run default

simonpintarelli:
cscs-ci run default

mtaillefumier:
The results differ between MPI=1 and MPI \neq 1. It is coming from the parallel fft and some strange interqaction between vdwxc and sirius slab decomposition.

mtaillefumier:
cscs-ci run default

mtaillefumier:
cscs-ci run default

mtaillefumier:
cscs-ci run default

src/context/simulation_context.cpp
```diff
@@ -690,20 +691,32 @@ Simulation_context::print_info(std::ostream& out__) const
         rte::ostream os(out__, "info");
         int i{1};
         os << std::endl << "XC functionals" << std::endl << hbar(14, '=') << std::endl;
+        std::vector<double> weight(xc_functionals().size(), 1.0);
+
+        /* Check if the weights are given and if the number of elements match the number of functionals. If empty, then default to 1.0
+         */
+        if (xc_functionals_weight().size() != 0) {
```
toxa81, line 703: Is `if (xc_functionals_weight().size() != 0) {` check redundant?

src/context/simulation_parameters.hpp
```diff
@@ -503,6 +503,12 @@ class Simulation_parameters
         return cfg().parameters().xc_functionals();
     }
 
+    std::vector<double>
```
toxa81: can be `inline auto`

src/potential/potential.cpp
```diff
@@ -65,12 +65,27 @@ Potential::Potential(Simulation_context& ctx__)
     }
 
     /* create list of XC functionals */
+
+    std::vector<double> weight(ctx_.xc_functionals().size(), 1.0);
+
+    /* Check if the weights are given and if the number of elements match the number of functionals. If empty, then default to 1.0
+     */
+    if (ctx_.xc_functionals_weight().size() != 0) {
```
toxa81: is `if (ctx_.xc_functionals_weight().size() != 0) {` check redundant?

mtaillefumier: See comment above. I can maybe try to merge these checks in one check only.

src/context/simulation_context.cpp
mtaillefumier: Actually this one is needed if we want to have a default behavior. SIRIUS will throw an error only if the weights are given and the number of weight does not match the number of functional. If the weights are not explicitly given then we default to 1.0 for each functional.

src/context/simulation_parameters.hpp
mtaillefumier: Ok let me try this.

mtaillefumier:
There is a bug in the API as well. When a field is declared object, SIRIUS still wants to have the "default" field defined which results in throwing an error.

src/context/simulation_context.cpp
toxa81: clear

src/potential/potential.cpp
toxa81: yes, probably can be fused into one `if` using C's short-circuit evaluation of `if` statements

simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

src/geometry/stress.cpp
```diff
@@ -362,8 +380,43 @@ Stress::calc_stress_xc()
         stress_xc_ += t;
     }
 
-    symmetrize_stress_tensor(ctx_.unit_cell().symmetry(), stress_xc_);
+    // compute the contributions of the stress tensor originating from non-local corrections
+    if (ctx_.full_potential()) {
+        potential_.xc_vdw_stress<false>(density_);
```
toxa81: There is no stress code for full-potential. Maybe the switch is for non-local core correction?

toxa81 (changes requested):
I found one suspicious place for stress tensor (full-potential switch)

simonpintarelli:
cscs-ci run default

src/geometry/stress.cpp
mtaillefumier: This is correct. I forgot about it. I should only call it when we use pseudo-potentials. I will fix it ASAP

mtaillefumier: the template parameter is the same than in
```
 template <bool add_pseudo_core__>
void
Potential::xc(Density const& density__)
```

mtaillefumier:
cscs-ci run default

src/potential/potential.cpp
mtaillefumier: Sorry I forgot to include this change. It is done now

src/context/simulation_parameters.hpp
mtaillefumier: Also included now. Thanks for the suggestion

toxa81:
hmm.. in the base container it's vdw-0.5.0.. why build_and_test_cpu finds 0.4.0

mtaillefumier:
Maybe I did a mistake when updating the docker file. I will check it

toxa81:
I checked your build, it was actually fine. I triggered it once again https://github.com/electronic-structure/SIRIUS/actions/runs/20426901540/job/58689018912 but I don't think it will do something different

mtaillefumier:
compilation passes. I only need to update the last two tests as dftd3 and dftd4 stress tensors were changed.

src/geometry/stress.cpp
```diff
@@ -362,8 +380,43 @@ Stress::calc_stress_xc()
         stress_xc_ += t;
     }
 
-    symmetrize_stress_tensor(ctx_.unit_cell().symmetry(), stress_xc_);
+    // compute the contributions of the stress tensor originating from non-local corrections
+    if (ctx_.full_potential()) {
```
toxa81: It can't be `if (ctx_.full_potential())`. Stress tensor only implemented for the pseudo-potential.

src/geometry/stress.cpp
toxa81: you probably just need to keep `potential_.xc_vdw_stress<true>(density_);`

mtaillefumier:
@toxa81 I can not find the last change request. Would it be possible to check again so that I include these changes before merging.

toxa81:
cscs-ci run default

toxa81 (approved):
LGTM

simonpintarelli:
Note: I'll rebase / cherry pick once the CI has been fixed the other PR (1113)

PR #1084, opened 2025-07-15
CI: fix cpus per task on eiger

simonpintarelli:
CI tests on eiger are running into wall time limit due to another change in slurm what previously was
`SRUN_CPUS_PER_TASK` seems to be again `SLRUM_CPUS_PER_TASK` ... 

Using both variables for now.


simonpintarelli:
cscs-ci run default

simonpintarelli:
cscs-ci run default

PR #1086, opened 2025-07-22
[ci/cd] build base gcc container for github

electronic-structure:


toxa81:
This has to be merged in order to be tested.

PR #1089, opened 2025-08-28
Elpa api change

simonpintarelli:
Needs to be merged before CI update to spack v1.0 (which will use elpa 2025.1)

From the elpa manual (https://elpa.mpcdf.mpg.de/userguide/archive/elpa_userguide_2024_05.pdf, page 35)
>Important note for the GPU users
The overloaded convenience functions, like eigenvalues() can only be used if the data has
been allocated on the host. If the data has been allocated on the GPU device, an automatic
destinction of datatypes is not possible and one has to use the explicit functions specifying
the datatype, e.g. eigenvalues_double()

It appears to me one could have used the overloaded functions already pre 2025.


simonpintarelli:
cscs-ci run default

PR #1094, opened 2025-09-30
Improve DLA-Future integration

RMeli:
DLA-Future was always initialized when SIRIUS was compiled with DLA-Future support (like ELPA). However, this can lead to issues when users are unaware of the underlying `pika` runtime and associated overheads.

This PR improves the integration by initializing DLA-Future on-demand and creating DLA-Future grids only when needed. The grids are nonetheless cached and freed only when the program is finalized.

This PR also adds ScaLAPACK requirements to Spack when SIRIUS is compiled with ELPA and/or DLA-Future. This requirement for the latter can be lifted by integrating `potrf` and `porti` from DLA-Future, which will be addressed in the future.


RMeli:
cscs-ci run default

mtaillefumier (approved):
Thanks. The code is simpler.

abussy (approved):
Not really an expert on this, but it runs fine when tested in a uenv.

simonpintarelli (approved):
lgtm

PR #1091, opened 2025-09-04
Docker ci

simonpintarelli:
Update docker ci to spack 1.0

TODO: bring back q-e-sirius build in a follow-up PR

PR #1095, opened 2025-09-30
Fix position of Spack compiler in CI

RMeli:


RMeli:
Closing; there are other failures and I initially missed https://github.com/electronic-structure/SIRIUS/pull/1091.

PR #1097, opened 2025-10-06
build qe

simonpintarelli:
add missing QE build to CI


toxa81:
Cmake must be part of the base image. It was commented and I thought there was a reason for that. Now there is a reason to bring it back :) Or at least, install as a step before building qe

simonpintarelli:
> Cmake must be part of the base image. It was commented and I thought there was a reason for that. Now there is a reason to bring it back :) Or at least, install as a step before building qe


Cmake from the system was too old.. `spack load cmake` might help 🤞🏻

simonpintarelli:
@toxa81 it shouldn't be necessary to install cmake manually, spack installed it as dependency of sirius, it just needs to be loaded correctly

toxa81:
cscs-ci run default

PR #1101, opened 2025-10-22
[ci/cd] compile with clang

toxa81:


.github/workflows/build.yml
```diff
@@ -37,3 +37,13 @@ jobs:
...
           make
+  build_and_test_clang:
+    runs-on: ubuntu-latest
+    container: electronicstructure/sirius:latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build SIRIUS
+        run: |
+          spack -e /sirius-env-clang develop -p ${GITHUB_WORKSPACE} sirius@develop~openmp
+          spack -e /sirius-env-clang concretize -f
```
simonpintarelli, line 48: I wanted to complain about the re-concretization.. Ideally it would be setup as develop spec in the env during the build of the base container. I think the main issue is that ${GITHUB_WORKSPACE} isn't known during the previous stage.

toxa81: But why do we need the ${GITHUB_WORKSPACE} in the previous stage? The source code is added here and now and then compilation is performed. What I am missing?

simonpintarelli: If the value of GITHUB_WORKSPACE were known previously, one could have setup a spack environment with sirius as develop spec and then call `spack concretize`, `spack install --only=dependencies`. Then we would only have to call `spack install` in this stage, without the risk of concretization changes / recompilation of dependencies.

simonpintarelli (approved):
lgtm

PR #1104, opened 2025-11-18
[enhancement] use k-independent muffin-tin part of the Hamiltonian

toxa81:
Muffin-tin Hamiltonian and potential matrix elements are k-point independent and can be precomputed once when H0 is created.


toxa81:
cscs-ci run default

mtaillefumier (approved):

PR #1108, opened 2025-12-01
Added # to the printf statement to accomade CP2K

mtaillefumier:
CP2K echoes input files using the standard output which conflicts with SIRIUS initialization message printed in the standard output. Added a # in front of the message so that the message is considered as a comment by CP2K.

https://github.com/cp2k/cp2k/issues/4564#issuecomment-3586844886

We may need a minor revision.

### DIFF
--- a/src/sirius.hpp
+++ b/src/sirius.hpp
@@ -91,10 +91,10 @@ initialize(bool call_mpi_init__ = true)
 #endif
 
     if (mpi::Communicator::world().rank() == 0) {
-        std::printf("SIRIUS %i.%i.%i, git hash: %s\n", major_version(), minor_version(), revision(),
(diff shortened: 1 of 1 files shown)


oschuett:
Thanks for taking this on Mathieu!

Personally, I'd really prefer a way to prevent SIRIUS from writing its banner all together.
Instead CP2K could print the version infos at a more appropriate place and using its own [IO facilities](https://www.cp2k.org/dev:printkey).

mtaillefumier:
I will ask if we can simply remove this banner altogether.

toxa81 (approved):
LGTM. I'm hesitant to remove the banner completely because I find it helpful.

PR #1111, opened 2025-12-09
v7.10.0

simonpintarelli:
bump version


toxa81:
@simonpintarelli two files are affected, second is the doxygen.cfg which also contains a version

PR #1112, opened 2025-12-22
[fix] mixer inner product

toxa81:
- "inner product" in Mixer is defined as a sum of a Hartree energy of a density residual + 1/pi of the inner product of magnetization components residuals
- "size" function in Mixer is removed as unnecessary
- `density_residual_hartree_energy` is removed as unnecessary

Because the RMS contribution is now computed from density and magnetization residuals, some reference values in tests changed a bit.


toxa81:
cscs-ci run default

simonpintarelli:
cscs-ci run default

mtaillefumier:
I do not understand why test 30 behaves erratically.

mtaillefumier (approved):
LGTM

toxa81:
cscs-ci run default

PR #1117, opened 2026-01-10
[fix] allocation of hmt on GPUs

toxa81:
Change the logic of allocation of hmt arrays. For some reason, it was producing wrong memory access on GPU for the local build.


toxa81:
cscs-ci run default

simonpintarelli (approved):
lgtm

PR #1120, opened 2026-01-15
[feature] add report for power measurments

toxa81:
* Only first rank of each node reads pm counters, and then they sum the total values between nodes. 
* Adjustments for 4 GPU cards


toxa81:
cscs-ci run default

PR #1131, opened 2026-02-13
Fix ROCm<6 compilation and unify device UUID formatting

alephpiece:
`hipDeviceProp_t.uuid` is defined on ROCm 6.0 and later, which causes compilation to fail on older ROCm versions (e.g., ROCm 5.x). This PR addresses this backward compatibility issue and fixes a string formatting bug in the `get_uuid()` function.

- Switched to using `hipUUID` and the `DeviceGetUuid` API directly for the ROCm backend, safely resolving the build issue for ROCm < 6.0.
- Unified the string formatting `for` loop. By abstracting the byte array into a constant reference (`auto const& uuid_bytes`), both CUDA and ROCm backends now share the exact same logic.

### DIFF
--- a/src/core/acc/acc.hpp
+++ b/src/core/acc/acc.hpp
@@ -336,25 +336,21 @@ get_uuid(int device_id__)
 {
 #if defined(SIRIUS_CUDA)
     cudaDeviceProp devprop;
+    CALL_DEVICE_API(GetDeviceProperties, (&devprop, device_id__));
+    auto const& uuid_bytes = devprop.uuid.bytes;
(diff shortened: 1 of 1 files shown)


toxa81:
cscs-ci run default

toxa81:
Dear @alephpiece 
Thanks for the PR. I trigger the build, but due to minor mistakes it won't compile. I think you need to safeguard with if-def statements cuda/rocm related variables so they don't appear in cpu-only build

alephpiece:
Thanks for the review! I've added back the safeguard.

toxa81:
cscs-ci run default

PR #1134, opened 2026-03-16
[fix] workaround to force GSL to link to openblas

toxa81:


toxa81:
cscs-ci run default

AdhocMan:
It looks like SIRIUS actually never directly calls `find_package(BLAS)` , so this solution probably depends on `BLAS_LIBRARIES` being populated by another module.
The `FindGSL` module defines the target as 
```cmake
    set_target_properties( GSL::gsl PROPERTIES
      IMPORTED_LOCATION                 "${GSL_LIBRARY}"
      INTERFACE_INCLUDE_DIRECTORIES     "${GSL_INCLUDE_DIRS}"
      IMPORTED_LINK_INTERFACE_LANGUAGES "C"
      INTERFACE_LINK_LIBRARIES          GSL::gslcblas )
```
So I think using
```cmake
set_target_properties(GSL::gsl PROPERTIES
    INTERFACE_LINK_LIBRARIES ""
)
```
after the call to `find_package(GSL)` might be a more robust solution.

toxa81:
Do we actually need to call `find_package(BLAS)` from sirius cmake?

toxa81:
I confirm, this
```
set_target_properties(GSL::gsl PROPERTIES
    INTERFACE_LINK_LIBRARIES ""
)
```
also works. No gslcblas is linked

AdhocMan:
> Do we actually need to call `find_package(BLAS)` from sirius cmake?

No, but you'd be dependent on behavior of other packages that is not guaranteed. 

Btw, I think you'd also have to add the fix to 
https://github.com/electronic-structure/SIRIUS/blob/develop/cmake/sirius_cxxConfig.cmake.in#L35
with a guard to check if `GSL::gsl` exists. This should avoid having that issue with downstream dependencies that also use CMake.

toxa81:
cscs-ci run default

AdhocMan (approved):
lgtm

PR #1139, opened 2026-04-07
Correction from DFTD3 to DFTD4 in the API

mrm24:
Changed some DFTD3 in the DFTD4 API to DFTD3.


mtaillefumier (commented):
Thanks for spotting this. It should be resolved with #1147

Growl1234:
I think this can be merged as well since it contains fixing another typo I did not noticed.

simonpintarelli:
cscs-ci run default

mtaillefumier (approved):

PR #1140, opened 2026-04-09
Fix to IBS force.cpp

mrm24:
For LAPW+lo IBS is giving NaN in the case there in some grid configurations (if more than one process is assigned to the k-point group). See the following output of exciting with SIRIUS:
image
This patch solves it by initializing h1 and o1 (the derivatives) to 0 inside the Cartesian axis loop. The result looks like:
Screenshot from 2026-04-09 14-29-07

### DIFF
--- a/src/geometry/force.cpp
+++ b/src/geometry/force.cpp
@@ -846,6 +846,10 @@ Force::add_ibs_force(K_point<double>* kp__, Hamiltonian_k<double>& Hk__, mdarray
         }
 
         for (int x = 0; x < 3; x++) {
+
+            o1.zero();
+            h1.zero();
+            
             for (int igk_col = 0; igk_col < kp__->num_gkvec_col(); igk_col++) { // loop over columns
                 auto gvec_col = kp__->gkvec_col().gvec(gvec_index_t::local(igk_col));
                 for (int igk_row = 0; igk_row < kp__->num_gkvec_row(); igk_row++) { // loop over rows


toxa81:
Dear @mrm24
thank you for submitting the fix! If possible, can you please share your input file and MPI grid layout (you can send it to me to antonk@cscs.ch). When time allows, I need to have a look at them. Why I'm concerned: from the code zeroing of h1 and o1 is not required. Each element of those auxiliary matrices will be set explicitly.  So while your fix is working, something is still not right

toxa81:
cscs-ci run default

src/geometry/force.cpp
```diff
@@ -846,6 +846,10 @@ Force::add_ibs_force(K_point<double>* kp__, Hamiltonian_k<double>& Hk__, mdarray
         }
 
         for (int x = 0; x < 3; x++) {
+
```
toxa81, line 849: I found the issue. It's the local orbital - local orbital part of the matrix has to be zeroed, because it doesn't contribute. Correct code is:
```
            /* zero lo-lo block */
            for (int icol = 0; icol < kp__->num_lo_col(); icol++) {
                for (int irow = 0; irow < kp__->num_lo_row(); irow++) {
                    h1(irow + kp__->num_gkvec_row(), icol + kp__->num_gkvec_col()) = la::constant<std::complex<double>>::zero();
                    o1(irow + kp__->num_gkvec_row(), icol + kp__->num_gkvec_col()) = la::constant<std::complex<double>>::zero();
                }
            }

```

toxa81 (changes requested):

toxa81:
fixed in a different MR

PR #1142, opened 2026-04-19
Update GH200 CI, cuda13

simonpintarelli:
- cuda13
- latest spack-packages
- add local recipes for dftd4 (and deps) with shared library, the static libraries of dftd4 caused a linker error with gfortran


simonpintarelli:
cscs-ci run default

PR #1147, opened 2026-04-24
Fix typo

Growl1234:
There are two places writing "dftd3" in dftd4 section (?)


simonpintarelli:
Thanks! I'll add it to the `7.11.1` patch.

PR #1155, opened 2026-06-15
CMake: Set SOVERSION for SIRIUS shared libraries

Growl1234:
Set VERSION and SOVERSION for SIRIUS shared libraries, so that installed shared objects carry proper SONAMEs and versioned symlinks (`libsirius.so.7.11.1`, `libsirius.so.7`, `libsirius.so`). This improves downstream packaging and makes ABI/runtime dependencies easier to track for distributions and external consumers.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -93,7 +93,10 @@ add_library(sirius_cxx "${_SOURCES};${_CUSOURCES};")
 target_compile_features(sirius_cxx PUBLIC cxx_std_17)
 target_compile_features(sirius_cxx PUBLIC c_std_99)
 target_compile_features(sirius_cxx PUBLIC cuda_std_17)
-set_target_properties(sirius_cxx PROPERTIES POSITION_INDEPENDENT_CODE ON)
+set_target_properties(sirius_cxx PROPERTIES
+  VERSION ${PROJECT_VERSION}
+  SOVERSION ${PROJECT_VERSION_MAJOR}
+  POSITION_INDEPENDENT_CODE ON)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -93,7 +93,10 @@ add_library(sirius_cxx "${_SOURCES};${_CUSOURCES};")
 target_compile_features(sirius_cxx PUBLIC cxx_std_17)
 target_compile_features(sirius_cxx PUBLIC c_std_99)
 target_compile_features(sirius_cxx PUBLIC cuda_std_17)
-set_target_properties(sirius_cxx PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
mtaillefumier, line 96: Please do not remove this option. Forcing it does not cost anything.

mtaillefumier (changes requested):

src/CMakeLists.txt
Growl1234: Hmm `POSITION_INDEPENDENT_CODE ON` is not removed and the real change here is:
```patch
- set_target_properties(sirius_cxx PROPERTIES POSITION_INDEPENDENT_CODE ON)
+ set_target_properties(sirius_cxx PROPERTIES
+  VERSION ${PROJECT_VERSION}
+  SOVERSION ${PROJECT_VERSION_MAJOR}
+  POSITION_INDEPENDENT_CODE ON)
```
The `set_target_properties()` call is just split into multiple lines for readability.

mtaillefumier (approved):

PR #1156, opened 2026-06-15
[fix] print banner only when verbosity is set

toxa81:


mtaillefumier (approved):

mtaillefumier:
I will merge it as soon as I have the green flag from the CI/CD.

PR #1160, opened 2026-07-03
CMake: Preserve ScaLAPACK link interfaces for static builds

Growl1234:
This reworks the linear-algebra dependency setup so that ScaLAPACK keeps its complete transitive link interface instead of being flattened into a manually ordered library list.

* Prefer a package-provided ScaLAPACK target, then a `pkg-config --static` target, before falling back to a user-provided ordered library list.
* Link `sirius::scalapack` before its MPI and LAPACK dependencies, which is required when ScaLAPACK and/or LAPACK are static archives.
* Use oneMKL's `MKL::MKL_SCALAPACK` target when ScaLAPACK support is enabled, allowing `MKLConfig.cmake` to provide the matching BLACS libraries and required static-link group options.
* Detect Open MPI, MPICH, and Intel MPI explicitly rather than treating every non-Open-MPI implementation as Intel MPI. (MPICH is compatible with Intel MPI wrappers)
* Preserve the imported `GSL::gsl` usage requirements instead of clearing its CBLAS dependency.
* Make the installed `sirius_cxx` package self-contained by including `CMakeFindDependencyMacro` and restoring the oneMKL configuration needed by downstream consumers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,7 +133,6 @@ endif()
 find_package(MPI REQUIRED)
 #set(GSL_CBLAS_LIBRARY ${BLAS_LIBRARIES} CACHE STRING "" FORCE)
 find_package(GSL REQUIRED)
-set_target_properties(GSL::gsl PROPERTIES INTERFACE_LINK_LIBRARIES "")
 find_package(LibXC 4.0.0 REQUIRED)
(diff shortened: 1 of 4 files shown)


mtaillefumier (approved):

toxa81:
cscs-ci run default

PR #1162, opened 2026-07-10
CMake: Remove cmake_minimum_required from Config

Growl1234:
Setting `cmake_minimum_required()` in `sirius_cxxConfig.cmake.in` will override the policy settings inherited from downstream projects, such as `CMP0144`.


toxa81:
cscs-ci run default

simonpintarelli (approved):
thanks! lgtm

PR #1164, opened 2026-07-14
[refactor] constrained lda+u implementation

toxa81:
Important changes:
- Logically group Hubbard constraint parameters in the `hubbard.constraint` section of the input; this requires corresponding changes to input files.
- `hubbard_constrained_calculation` is now determined by the presence of elements in `hubbard.constraint.local`.
- `hubbard_subspace_method` is renamed to `subspace_method` within the `hubbard` input section.
- `hubbard.constraints_enabled` is removed as redundant.
- `input_schema.json` is updated, and `config.hpp` is regenerated.
- `Hubbard_matrix` now returns `num_atomic_levels()` for local corrections and constraints, and `num_nonlocal()` for the number of non-local corrections.
- Access methods that return full objects from `Hubbard_matrix` are removed.
- `apply_constraints_` is renamed to `active_constraints_`.
- `apply_constraint()` is renamed to `apply_constraints()`.
- `atomic_orbitals()` is renamed to `atomic_orbital()`.
- A const reference to `Hubbard_matrix` is now stored in `U_operator`.
- `num_atomic_levels_` is now a member of the `Hubbard_matrix` class.
- A redundant double loop over constraint indices in `hubbard_matrix.cpp` is removed.
- Fix a bug in `hubbard_matrix.hpp` where constraint multipliers were copied by iterating over `nonlocal().size()`.
- PoC: add constrained potential to LAPW matrix elements


toxa81:
@mtaillefumier 

https://github.com/electronic-structure/SIRIUS/blob/develop/src/hubbard/hubbard_matrix.hpp#L306

is probably the reason why test30 doesn't work in this PR. nonlocal.size() is zero and multipliers_constraints were not copied

mtaillefumier:
> @mtaillefumier
> 
> https://github.com/electronic-structure/SIRIUS/blob/develop/src/hubbard/hubbard_matrix.hpp#L306
> 
> is probably the reason why test30 doesn't work in this PR. nonlocal.size() is zero and multipliers_constraints were not copied

indeed as far as I remember the constrained hubbard only apply to local hubbard so the copy will not happen at all. that's a nice catch.

mtaillefumier (approved):
Strange, it is failing in one ci/cd not with clang

mtaillefumier:
The constrained hubbard implementation is clearly broken somewhere. I should investigate. My suggestion is to turn off the test, open an issue assign it to me and merge the PR because many of these changes are valuable. 

As far as constrained hubbard goes I remain convinced that the theory is questionable to begin with. history will repeat itself (SO interaction only checked by the specialists 6 years after its implementation), I come to the conclusion that this functionality will be used in four years.

toxa81:
test30 will converge if I switch off full orthogonalization; so it has something to do with the subset of fully orthogonalized orbitals used in calculation 
Update: with beta_mixing: 0.25 also full_orthogonalization case converges

toxa81:
>Strange, it is failing in one ci/cd not with clang
clang pipeline is only used to check compilation. No tests are run there

toxa81:
cscs-ci run default

PR #1165, opened 2026-07-17
[refactor] Move plane-wave coefficients to Hamiltonian0

toxa81:
Major changes:
- Get rid of the _quck-fix_ API function to construct Hamiltonian. Now this is done by the proper function `sirius_create_hamiltonian()`
- Some constants are moved to constants.hpp
- LAPW matrix elements of effective potential, relative mass, and its square are now computed in Hamiltonian0 where they are consumed; setter function are removed, because these objects can be recomputed from the potential and step function
- Local_operator and Hamitonial0 are now using common functions defined in interstitial_functions.hpp

q-e-sirius is tested for API change: https://github.com/electronic-structure/q-e-sirius/pull/96

Verification tests are passing, but the compilation of dev branch of q-e-sirius is failing because of the API change.


toxa81:
cscs-ci run default

mtaillefumier (approved):
Question, are physical constant values taken from NIST ? I think we should state this clearly to avoid argumentation why results may differ.

toxa81:
> Question, are physical constant values taken from NIST ? I think we should state this clearly to avoid argumentation why results may differ.

I don't recall but 137.035999139 is what is commonly used everywhere (https://en.wikipedia.org/wiki/Fine-structure_constant, CODATA-2014). However, for the published NIST data with ScRLDA functional, I wasn't able to reproduce the total energy values. Then I dig deeper and found that they have used 137.0359895 (CODATA-1986)

mtaillefumier:
> > Question, are physical constant values taken from NIST ? I think we should state this clearly to avoid argumentation why results may differ.
> 
> I don't recall but 137.035999139 is what is commonly used everywhere (https://en.wikipedia.org/wiki/Fine-structure_constant, CODATA-2014). However, for the published NIST data with ScRLDA functional, I wasn't able to reproduce the total energy values. Then I dig deeper and found that they have used 137.0359895 (CODATA-1986)

Yes, a small difference in the constant values can be problematic. So just in case, it is good to indicate where the value comes from as you did. It is really a matter of traceability and should not block merging the PR.

PR #1166, opened 2026-08-03
Describe sirius_cxx as a dependent library for shared builds

Growl1234:
Describe `libsirius_cxx` as an implementation dependency of the imported `sirius::sirius` target when SIRIUS is built with shared libraries.

This allows consumers to link only against `sirius::sirius` without loading the full `sirius_cxx` dependency interface, which can otherwise rediscover conflicting ELPA or DFT-D3 providers and leak C++ usage requirements into Fortran consumers.

Static builds retain the existing `find_dependency(sirius_cxx CONFIG)` behavior because they require the complete transitive link closure.


simonpintarelli:
cscs-ci run default

simonpintarelli (approved):
thanks! lgtm

simonpintarelli:
(for the record)
There is an unrelated error in the CPU tests on eiger:
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/7415069138667150/4626796664769983/-/jobs/15693248400/viewer#L15762

Magnetisation is off in test09

PR #1167, opened 2026-08-18
Clean-up API (unnecessary Julia related functions)

abussy:
Cleaned-up the SIRIUS API to remove functions that where added for a fine grained DFTK integration. We have moved on since, so I only left functions required in `SIRIUS.jl`, which allows running SCFs from Julia.


toxa81:
@abussy cpu build fails at the moment

abussy:
Forgot to re-generate the `sirius.f90` file.

toxa81:
cscs-ci run default

